PR #22, opened 2021-04-30
Create stepper component

przemyslawjanpietrzak:


lib/js/atoms/Stepper.vue
```diff
@@ -0,0 +1,117 @@
+<template>
+	<div class="stepper">
+		<div v-for="(step, index) in steps" :key="index" class="stepper__item">
+			<div
+				v-if="index !== 0"
+				class="stepper__separator"
+				:class="{ '-active': step.isActive }"
+			></div>
+			<div class="stepper__icon" :class="{ '-active': step.isActive }">
```
kvas-damian, line 9: Is it intended that multiple steps are active at the same time? 
Maybe we should consider having only one step active, all steps before that step should be also marked green. Such encapsulation avoids leaking isActive logic to the application

lib/js/atoms/Stepper.vue
```diff
@@ -0,0 +1,117 @@
...
+				<icon :icon="ICONS[step.iconKey]" :size="ICON_SIZES.X_SMALL"></icon>
```
kvas-damian, line 10: On designs I see different icon for steps that are finished. How do we want to address that?

lib/js/atoms/Stepper.vue
```diff
@@ -0,0 +1,117 @@
...
+		}
+	}
+
+	&__icon {
+		align-self: center;
+		background-color: $color-minor-background;
+		border-radius: 100%;
+		color: $color-minor;
+		display: flex;
+		height: 32px;
```
kvas-damian, line 51: variables

lib/js/atoms/Stepper.vue
```diff
@@ -0,0 +1,117 @@
...
+		@include textBold();
+
+		color: $color-minor;
+
+		&.-active {
+			color: $color-primary;
+		}
+	}
+
+	&__icon {
```
kvas-damian, line 45: do we need additional wrapper on icon?
Maybe we could apply styles on `<wnl-icon class="stepper__icon">`? This way you could avoid setting div size

lib/js/atoms/Stepper.vue
```diff
@@ -0,0 +1,117 @@
...
+		transition: background-color 0.33s;
+
+		&.-active {
+			background-color: $color-primary-background;
+			color: $color-primary;
+		}
+	}
+
+	&__separator {
+		background-color: #e6e8ee;
```
kvas-damian, line 64: variables

lib/js/atoms/Stepper.vue
```diff
@@ -0,0 +1,117 @@
...
+		border-radius: 2.5px;
```
kvas-damian, line 65: please discuss with UX if we really want to use decimals here

lib/js/atoms/Stepper.vue
```diff
@@ -0,0 +1,117 @@
...
+		height: 2px;
+		margin: 0 $space-xs;
+		transition: background-color 0.33s;
```
kvas-damian, line 68: I see `0.33s` multiple times. It's a nice candiadate for a variable. Can be local for now...

lib/js/atoms/Stepper.vue
```diff
@@ -0,0 +1,117 @@
...
+		width: 32px;
```
kvas-damian, line 69: it shouldn't be fixed width. We can define mix and max, but definitely it should be adjusting based on the number of steps and window size

lib/js/atoms/Stepper.vue
```diff
@@ -0,0 +1,117 @@
...
+	props: {
+		steps: {
+			type: Array as Prop<Array<Step>>,
+			required: true,
+			validate: (steps: Array<Step>) => {
+				return steps.every((step) => {
+					if (typeof step.label !== 'string') {
+						return false;
+					}
+					if (typeof step.iconKey !== 'string') {
```
kvas-damian, line 100: probably we should validate if Icon with this name exists

lib/js/atoms/Stepper.vue
```diff
@@ -0,0 +1,117 @@
...
+export default {
+	name: 'Stepper',
+	components: { Icon },
+	props: {
+		steps: {
+			type: Array as Prop<Array<Step>>,
+			required: true,
+			validate: (steps: Array<Step>) => {
+				return steps.every((step) => {
+					if (typeof step.label !== 'string') {
```
kvas-damian, line 97: won't TS validate that? Do we need to do that manually?

lib/js/atoms/Stepper.vue
```diff
@@ -0,0 +1,117 @@
...
+@import '../../styles/settings/media-queries';
+@import '../../styles/settings/typography';
+
+.stepper {
+	display: flex;
+
+	&__item {
+		align-items: center;
+		display: flex;
+		justify-content: center;
```
kvas-damian, line 31: On mobile label is under the icon on designs

lib/js/atoms/Stepper.vue
przemyslawjanpietrzak: I thought about it, however I choose this solution

motivation:
 - this is DS component so should be as reusable and logic agnostic as possible
 - on the application level we gonna to write wrapper for applying that logic (if stepper will be only application component this 2 components should be merged)

lib/js/atoms/Stepper.vue
przemyslawjanpietrzak: as above, in my opinion it's application specific logic, so it should not be in DS

lib/js/atoms/Stepper.vue
przemyslawjanpietrzak: Agree, but let's keep in `styles/settings`, ok?

lib/js/atoms/Stepper.vue
przemyslawjanpietrzak: we do that, but in a child component

lib/js/atoms/Stepper.vue
przemyslawjanpietrzak: TS doesn't works on runtime, validator works. Also TS doesn't check props applied to rendered component.

I'm not fan of validators (e.g. react with TS doesn't validate props in runtime, however JSX allows to make that check in compilation time.

lib/js/atoms/Stepper.vue
```diff
@@ -0,0 +1,157 @@
...
+			width: 151px;
+		}
+
+		&.-filled {
+			background-color: $color-primary;
+		}
+	}
+
+	&__ripple {
+		height: $step-icon-size;
```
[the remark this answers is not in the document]
kvas-damian: but then ripple is too small :(

lib/js/atoms/Stepper.vue
przemyslawjanpietrzak: Ewa is gonna to send us exact requirements, however we don't need it right now

lib/js/atoms/Stepper.types.ts
```diff
@@ -0,0 +1,7 @@
+export interface Step {
+	label: string;
+	iconKey: string;
```
kvas-damian, line 3: Passing `iconKey` can cause some trubble here...
You need to hardcode string, or hack the map of icons to get the key. What do you think about passing icon object instead? Link in `Button` component?

lib/js/atoms/Stepper.vue
kvas-damian: @przemyslawjanpietrzak any update here. I still see it's not fixed

lib/js/atoms/Stepper.vue
```diff
@@ -0,0 +1,157 @@
...
+	created() {
+		this.ICONS = ICONS;
+		this.ICON_SIZES = ICON_SIZES;
+	},
+	methods: {
+		onStepClick(step: Step) {
+			if (!step.isClickable) {
+				return;
+			}
+			this.$emit('click', step.route);
```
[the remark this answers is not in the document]
kvas-damian: What about event name? Do we want to address that?

lib/js/atoms/Stepper.vue
```diff
@@ -0,0 +1,152 @@
...
+		margin-right: $space-xxs;
+		width: $step-icon-size;
+		transition: background-color $basic-transition-time;
+
+		&.-filled {
+			background-color: $color-primary-background;
+			color: $color-primary;
+
+			&:hover:not(.-disabled) {
+				background-color: $color-primary-background-hovered;
```
kvas-damian, line 74: This color looks different in Invision. Can we double-check that?

przemyslawjanpietrzak: UX has no proposition how colors should be calculated, they agreed to stay with mix of 16% firefly-black

PR #24, opened 2021-05-04
IT-3620 | Use vue-material-adapter

kvas-damian:


package.json
```diff
@@ -60,7 +60,9 @@
     "@fortawesome/pro-regular-svg-icons": "^5.15.3",
     "@fortawesome/pro-solid-svg-icons": "^5.15.3",
     "@fortawesome/vue-fontawesome": "^2.0.2",
+    "@vue/composition-api": "^1.0.0-rc.8",
```
rogatty, line 63: Will we need it in the apps, too?

package.json
```diff
@@ -60,7 +60,9 @@
...
     "vue": "^2.6.11",
+    "vue-material-adapter-legacy": "^0.17.3",
```
rogatty, line 65: Legacy version wasn't updated for a long time: https://github.com/pgbross/vue-material-adapter/blob/legacy/release-notes.md

It uses material-components-web v7.0.0 (https://github.com/pgbross/vue-material-adapter/issues/171) while there is already v11.0.0 (https://github.com/material-components/material-components-web/blob/master/CHANGELOG.md).

Are you sure this is the right way?

package.json
przemyslawjanpietrzak: This lib gonna allow to turn on `noImpliciteThis` flag in the marketplace

kvas-damian: Yes, if we install this plugin in DS it'll be avaialable in `wnl-platform` and `marketplace-frontend`. Although, it's not recommended to use libraries/plugin that are imported by dependencies.

Summarizing, if you want to use it, you should install it as a direct dependency in your app, otherwise you don't need to know that we installed it.

package.json
kvas-damian: We aren't sure, but we believie it's the fastest way to achieve our goal: Material like forms.
As a backup plan we believe that ejecting `vue-material-adapter` and moving towards `material-components-web` should be relatively easy if necessary.

Discussion about this topic: https://bethinkteam.slack.com/archives/C01DAPQ63PD/p1620114175002500

PR #31, opened 2021-05-18
Create input component & install vuetify & add build step

przemyslawjanpietrzak:


lib/js/atoms/Input/Input.vue
```diff
@@ -0,0 +1,250 @@
...
+		background: $color-success-background !important;
+	}
+
+	.v-input--is-disabled .v-input__slot::before {
+		border-image: none !important;
+	}
+
+	&.-success {
+		.v-input__prepend-inner {
+			fill: $color-success;
```
kvas-damian, line 133: using `fill` will break duotone FA icons. We have to use `color` instead.

kvas-damian: Most likely you'll need to setup vue-svg-loader in storybook and library build to make it working

.postcssrc.js
```diff
@@ -0,0 +1,24 @@
...
+				if(selector.includes('select') || selector.includes('button') || selector.includes('textarea')) {
+					return prefixedSelector;
+				}
+
+
+				return selector;
+			}
+		}),
+	]
+}
```
kvas-damian, line 24:
```suggestion
}

```

.storybook/preview.js
```diff
@@ -77,3 +79,14 @@ export const parameters = {
...
+addDecorator(() => ({
+	components: {
+		App
+	},
+	template: `
+    <app>
+      <story/>
+    </app>
+    `,
+}));
```
kvas-damian, line 92:
```suggestion
}));

```

README.md
```diff
@@ -1,8 +1,5 @@
 # Bethink design system
 
-## Library info
-For quicker development during initial phase of Design System we decided to publish this repo as is - not as a [Vue library](https://cli.vuejs.org/guide/build-targets.html#library).
-
 ### Typescript
```
kvas-damian, line 3: We serve parsed code now, do we still need this section?

package.json
```diff
@@ -30,16 +32,20 @@
     "@storybook/vue": "^6.0.28",
     "@typescript-eslint/eslint-plugin": "^3.9.0",
     "@typescript-eslint/parser": "^3.9.0",
+    "@vue/cli-plugin-typescript": "^4.5.13",
     "@vue/cli-plugin-unit-jest": "^4.5.3",
+    "@vue/cli-service": "^4.5.13",
     "@vue/test-utils": "^1.1.3",
     "babel-eslint": "10.1.0",
+    "deepmerge": "^4.2.2",
```
kvas-damian, line 40: is it used?

README.md
kvas-damian: We need to export `initialize` and `initializeWithDomWatcher` from library to achieve that

lib/js/consts/vuetifyTheme.const.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const VUETIFY_THEME = {
+	themes: {
+		light: {
+			primary: THEME_COLORS.PRIMARY,
+			secondary: THEME_COLORS.SECONDARY,
+			error: FUNCTIONAL_COLORS.DANGER,
+			success: FUNCTIONAL_COLORS.SUCCESS,
+			warning: FUNCTIONAL_COLORS.WARNING,
+			anchor: 'none',
```
kvas-damian, line 11: > Invalid property value

przemyslawjanpietrzak: that was the point, we don't want to set color to all links in an application

README.md
przemyslawjanpietrzak: we do not, hover I suggest to keep it here until we make some webpack usage document on conflu

.postcssrc.js
```diff
@@ -0,0 +1,24 @@
...
+			transform: function (prefix, selector, prefixedSelector) {
+				if ((selector.startsWith('.v-application') || selector.startsWith('.theme--light.v-application')) && !selector.startsWith('.v-application--')) {
+					return prefixedSelector;
+				}
+
+				if(selector.includes('html')) {
+					return prefixedSelector;
+				}
+
+				if(selector.includes('select') || selector.includes('button') || selector.includes('textarea')) {
```
kvas-damian, line 15: these rules remove all styles that contain these keywords in the selector. We have to limit that

.postcssrc.js
```diff
@@ -0,0 +1,24 @@
+
+module.exports = {
+	plugins: [
+		require('postcss-prefix-selector')({
```
kvas-damian, line 4: It would be great to have the same rules applied to storybook, so we see changes before installing them in the real Application

.postcssrc.js
```diff
@@ -0,0 +1,24 @@
...
+			transform: function (prefix, selector, prefixedSelector) {
+				if ((selector.startsWith('.v-application') || selector.startsWith('.theme--light.v-application')) && !selector.startsWith('.v-application--')) {
+					return prefixedSelector;
+				}
+
+				if(selector.includes('html')) {
+					return prefixedSelector;
+				}
+
+				if(selector.includes('select') || selector.includes('textarea')) {
```
kvas-damian, line 15: these conditions are too general. It's very likely they will cause problems when we use these keywords in our selectores. Please convert them to somethinkg like:
* selector starts with X
* selector contains X preceded by space
* selector contains X preceded by comma

README.md
kvas-damian: this makes platform build faster and without hacks. IMO it's worth doing it the right way now.

.postcssrc.js
```diff
@@ -0,0 +1,23 @@
...
+			transform: function (prefix, selector, prefixedSelector) {
+				if ((selector.startsWith('.v-application') || selector.startsWith('.theme--light.v-application')) && !selector.startsWith('.v-application--')) {
+					return prefixedSelector;
+				}
+
+				if(selector === 'html') {
+					return prefixedSelector;
+				}
+
+				if(selector.match(/^select|textarea/) || selector.match(/( |,)(select|textarea)/) ) {
```
kvas-damian, line 15: https://regex101.com/

`/^select|textarea/` regex matches `xxx_textarea` string

kvas-damian, line 15:
```suggestion
				if(selector.match(/^(select|textarea)/) || selector.match(/( |,)(select|textarea)/) ) {
```

kvas-damian, line 15:
```suggestion
				if(selector.match(/^(select|textarea)/) || selector.match(/( |,)(select|textarea)/) ) {
```

PR #51, opened 2021-07-01
IT-3247 | create Badge component

przemyslawjanpietrzak:


lib/js/atoms/Badge/Badge.vue
```diff
@@ -0,0 +1,73 @@
...
+</template>
+
+<style lang="scss" scoped>
+@import '../../../styles/settings/colors';
+@import '../../../styles/settings/radiuses';
+@import '../../../styles/settings/typography';
+@import '../../../styles/settings/spacings';
+@import '../../../styles/atoms/icons';
+
+.badge {
```
rogatty, line 17: It conflicts with https://github.com/bethinkpl/wnl-platform/blob/1b5982c53778a140028e05135f157f523fd41696/resources/assets/js/components/global/Avatar.vue#L46-L51 and https://github.com/bethinkpl/wnl-platform/blob/1b5982c53778a140028e05135f157f523fd41696/resources/assets/styles/_mixins.scss#L77-L90.

lib/js/atoms/Badge/Badge.vue
```diff
@@ -0,0 +1,73 @@
...
+		align-items: center;
+		margin-right: $space-xxs;
+		height: $icon-xxs;
+		width: $icon-xxs;
+
+		& > img {
+			height: $icon-xxs;
+			width: $icon-xxs;
+		}
+	}
```
rogatty, line 37: This will break as soon as we introduce `BadgeWithIcon`. IMO this should be a slot without any icon-related styling. I'd name it `badge_media` and leave only the positioning. Media styles should come from above, in the slot.

lib/js/atoms/Badge/Badge.vue
```diff
@@ -0,0 +1,73 @@
...
+import { BADGE_COLORS } from './Badge.consts';
+
+export default {
+	name: 'Badge',
+	props: {
+		label: {
+			type: String,
+			required: true,
+		},
+		color: {
```
rogatty, line 61: I'd name it `backgroundColor`, see https://bethinkteam.slack.com/archives/CFBLF5RGA/p1625137653015700?thread_ts=1625127129.002700&cid=CFBLF5RGA

lib/js/atoms/Badge/Badge.vue
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	&__icon {
+		display: flex;
+		align-items: center;
+		margin-right: $space-xxs;
+		height: $icon-xxs;
+		width: $icon-xxs;
+
+		& > img {
```
kvas-damian, line 33: this won't work if we put `svg` into the slot or if the `img` will be wrapped by `div`

lib/js/atoms/Badge/Badge.vue
```diff
@@ -0,0 +1,73 @@
...
+			height: $icon-xxs;
+			width: $icon-xxs;
```
kvas-damian, line 35: `width: 100%` is more flexible, expecially if we decide to add size modifiers

lib/js/atoms/Badge/Badge.vue
kvas-damian: similar for the class name. It's very likely to clash with color modifiers in: https://github.com/bethinkpl/design-system/blob/eade93866eeddc4f3bd90bb08471dbfb8f52617e/lib/styles/modifiers/_colors.scss

lib/js/atoms/Badge/Badge.vue
przemyslawjanpietrzak: but this classes are scoped

przemyslawjanpietrzak: btw this is public repo

lib/js/atoms/Badge/Badge.vue
przemyslawjanpietrzak: @rogatty I don't see any advantages of `backgroundColor` compare to `color`

przemyslawjanpietrzak: @kvas-damian I though we decided to keep BEM over css utils

lib/js/atoms/Badge/Badge.vue
przemyslawjanpietrzak: will be removed

lib/js/atoms/Badge/Badge.vue
kvas-damian: `backgroundColor` allows us to add new color features to this component in future. On the other hand many UI libraries use `color` as a prop name for similar components: https://vuetifyjs.com/en/components/chips/#usage

The problem with className clash I was refering to isn't negating current solution. I was thinking what will happen if someone introduces `-minor` in modifiers file. That's why I suggest to avoid this class name.

rogatty: > @rogatty I don't see any advantages of `backgroundColor` compare to `color`

Component API is more obvious with `backgroundColor`, I see this as an advantage. Someone in the future could assume that `color` also affects the slot media, which isn't true. It's just the background we're coloring.

Unless we plan to synchronize font color with background color at this component's level. In that case `color` would be better 🙂

lib/js/atoms/Badge/Badge.vue
```diff
@@ -0,0 +1,68 @@
...
+	&.-minor {
+		background-color: $color-minor-background;
+	}
+
+	&__slot {
+		display: flex;
+		align-items: center;
+		margin-right: $space-xxs;
+		height: $icon-xxs;
+		width: $icon-xxs;
```
rogatty, line 31: I'm not sure if this is correct assumption but I'm fine with leaving it for now.

lib/js/atoms/Badge/Badge.vue
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	&__label {
+		@include textInfoS();
+		@include textBold();
+
+		color: $color-minor;
```
rogatty, line 38: Should be moved to `.-minor`, no?

lib/js/atoms/Badge/Badge.vue
przemyslawjanpietrzak: this minor requires 2 classes `&.-minor` so should have greater css score. But it will not protect for a `!important` syntax

lib/js/atoms/Badge/Badge.vue
przemyslawjanpietrzak: I don't know yet, there's no any other designs of this component

lib/js/atoms/Badge/Badge.vue
przemyslawjanpietrzak: I can rename it into `badgeMinor` however it will against aBEM

przemyslawjanpietrzak: Second idea it to don't use `-` prefix in not aBEM classes

rogatty: IMO the exact classname is of a less importance. It's an internal detail which can be changed at any time.

The important decision is the prop name as it's a public API. IMO:
- If we're going to synchronize background and text color at this component's level then the prop should be `color`
- If we're not, then the prop should be `backgroundColor` which will later allow us to add `textColor` etc.

przemyslawjanpietrzak: I'll choose first option ;)

rogatty: Let's ask UX what they choose 🙂

rogatty: During the meeting we've agreed that the icon color would be computed in `BadgeWithIcon` and not here (documented in https://bethinkteam.slack.com/archives/CFBLF5RGA/p1625137653015700). Computing text color here doesn't seem to be consistent with it. That's why `backgroundColor` IMO is simpler and safer.

przemyslawjanpietrzak: UX has no time in this sprint https://bethinkteam.slack.com/archives/CFBLF5RGA/p1625144844018000?thread_ts=1625127129.002700&cid=CFBLF5RGA . Let's keep it like this and wait for designs with other colors

rogatty (approved):

PR #52, opened 2021-07-06
IT-3247 | create Badge component & remove dead icon

przemyslawjanpietrzak:


lib/js/atoms/Badge/Badge.stories.mdx
```diff
@@ -0,0 +1,27 @@
...
+import Badge from './Badge'; import { BADGE_COLORS } from './Badge.consts';
+
+<Meta title='atoms/Badge' component={Badge} />
+
+# Badge
+
+<Story name='interactive'
+	   args={{
+		   label: 'Komentarz zespołu',
+		   iconUrl: 'https://lek.wiecejnizlek.pl/images/lek/logo-badge.svg',
```
rogatty, line 11: There is no `iconUrl` prop in `Badge` component.

rogatty (approved):
One minor thing and will LGTM

lib/js/atoms/Badge/Badge.stories.mdx
przemyslawjanpietrzak: And I don't pass this value to prop. I have to present this component for DR purses.

rogatty: Ah, I see now 👍 Expect some questions, though, because having a controlable argument suggests there is a prop like that.

PR #70, opened 2021-10-15
deploy to platform script

przemyslawjanpietrzak:


package.json
```diff
@@ -12,6 +12,7 @@
   },
   "scripts": {
     "build": "vue-cli-service build --target lib --name design-system lib/js/index.ts",
+    "deploy:platform": "yarn build && rm -rf ../wnl-platform/node_modules/design-system/dist && cp -r  dist ../wnl-platform/node_modules/design-system",
```
kvas-damian, line 15: initially I thought it's the right place for this script, but now I feel, we should move it to platform if possible.

Platform knows its dependencies and in which directory they live. @przemyslawjanpietrzak do you think it's possible?

PR #73, opened 2021-10-28
storybook-addon-pseudo-states PoC

kvas-damian:
example: https://smartdevs.pl/state/?path=/story/atoms-iconbutton--examples-multiple&globals=pseudo.hover:true


kvas-damian:
This addon has performance problems. Interesting alternative based on https://github.com/giuseppeg/postcss-pseudo-classes looks promising

PR #98, opened 2022-03-09
Colors parsers

damianmef:
+ All in one importer
+ New scss and json files with new raw colors and tokens
+ New views for colors and tokens
+ Little change in structure, "Basics" is now "Foundations", old colors are in "Deprecated" section


package.json
```diff
@@ -23,14 +23,16 @@
     "storybook": "start-storybook -p 6006",
     "storybook:build": "build-storybook -c .storybook -o public/storybook",
     "test": "jest",
-    "ts:check": "tsc --noEmit"
+    "ts:check": "tsc --noEmit",
+    "import": "npx ts-node lib/js/importers/RawColorsImporter.ts && npx ts-node lib/js/importers/TokensImporter.ts"
```
kvas-damian, line 27: We're importing colors when running this command for the first time. What do you think about renaming this script to `synchronizeColors`?

lib/js/styles/Colors/Colors.vue
```diff
@@ -0,0 +1,92 @@
...
+		<div class="sb__row">
+			<ol class="sb__list">
+				<li v-for="(colors, colorCategory) in rawColors" :key="colorCategory" class="sb__legend">
+					{{ colorCategory }}
+				</li>
+			</ol>
+		</div>
+		<div v-for="(colors, colorCategory) in rawColors" :key="colorCategory" class="sb__row">
+			<h2>{{ colorCategory }}</h2>
+			<div v-for="color in colors" :key="color.id" class='sb__table'>
```
kvas-damian, line 12: class name is not matching with tag name. Could you find a better name?

lib/js/styles/Colors/Colors.vue
```diff
@@ -0,0 +1,92 @@
...
+				<div class="sb__tableId">{{ color.label }}</div>
+				<div class="sb__tableValue">{{ color.value }}</div>
+				<div class="sb__tableColor">
+					<span class="sb__color" :style="{ background: color.value }"></span>
```
kvas-damian, line 16: why do we need two separate DOM nodes? Can we move styles to parent node?

lib/js/styles/Colors/Colors.vue
```diff
@@ -0,0 +1,92 @@
...
+
+		h2 {
+			margin-top: 0;
+		}
+	}
+
+	&__row {
+		border: 1px solid #f3f2f2;
+		padding: 20px  20px;
+		margin-bottom: 20px;
```
kvas-damian, line 37: we try to keep alphabetical order of style attributes

lib/js/styles/Colors/Colors.vue
```diff
@@ -0,0 +1,92 @@
...
+	}
+}
+</style>
+
+<script lang="ts">
+
+import rawColors from '../../../styles/settings/colors/_raw.json'
+
+export default {
+	name: 'Colors',
```
kvas-damian, line 85: this component is only used to document or DS, I'm not sure if we should put it into `lib` directory

lib/js/styles/Colors/Colors.vue
```diff
@@ -0,0 +1,92 @@
...
+	data(){
+		return {
+			rawColors
```
kvas-damian, line 88: to consider: maybe this component should be independent from colors source file. We can pass colors list as a property

lib/js/styles/ColorsThemes/ColorsThemes.vue
```diff
@@ -0,0 +1,50 @@
...
+</template>
+
+
+<script lang="ts">
+
+import themeWNL from '../../../styles/settings/colors/_raw.json'
+import themeBodywork from '../../../styles/settings/colors/_raw-bodywork.json'
+
+export default {
+	name: 'ColorsThemes',
```
kvas-damian, line 42: this component uses styles defined in another component, it's almost identical as `Colors.vue`. I think we should consider changes sugested in https://github.com/bethinkpl/design-system/pull/98/files#r822549545 to address this problems

lib/js/styles/Deprecated/Colors.stories.mdx
```diff
@@ -10,7 +10,6 @@ Color names generator can be found here: https://www.color-blindness.com/color-n
 
 ## Theme colors
 <ColorPalette>
-  <ColorItem title="Primary" colors={[THEME_COLORS.PRIMARY]} />
```
kvas-damian, line 13: is it intentionally removed? I'd keep it here until we fully migrate to new colors

lib/styles/modifiers/_colors.scss
```diff
@@ -27,7 +29,7 @@
 }
 
 .-fireflyBlack {
```
kvas-damian, line 31: IMO, it isn't enough to change color values here, we should also migrate class names. For this phase of migration I think we shouldn't touch this file

lib/styles/settings/_colors.scss
```diff
@@ -31,7 +31,7 @@ $color-ice-orange: #ffe6ca;
 $color-gigas-violet: #62378d;
 $color-deluge-violet: #7b569f;
 
-$color-total-black: #000;
+//$color-total-black: #000; //DEPRECATED
```
kvas-damian, line 34: why is it commented out? I believe all these colors defined in this file are deprecated

lib/styles/settings/colors/_tokens.scss
```diff
@@ -0,0 +1,304 @@
...
+$neutral-background-medium: --raw-gray-200;
+$neutral-background: --raw-gray-50;
+$neutral-background-hovered: --raw-gray-100;
+$neutral-background-focused: --raw-gray-100;
+$neutral-background-disabled: --raw-gray-50;
+$neutral-background-weak: --raw-white;
+$neutral-background-weak-hovered: --raw-gray-100;
+$neutral-background-weak-focused: --raw-gray-100;
+$neutral-background-weak-pressed: --raw-gray-200;
+$neutral-background-ghost: rgba(var(--raw-gray-700), 0);
```
kvas-damian, line 102: IMO, this will break. CSS `rgba()` method accepts `red, green, blue, alpha` parameters, not hex value

lib/js/importers/AllInOneImporter.ts
```diff
@@ -0,0 +1,189 @@
+const fsT = require('fs');
+const axios = require("axios");
+const tokensFilesConfig = JSON.parse(
+	fsT.readFileSync('lib/js/importers/configs/allInOneConfig.json', 'utf8')
+);
```
kvas-damian, line 5:
```suggestion
const tokensFilesConfig = require('./configs/allInOneConfig.json');
```

package.json
kvas-damian, line 27: these scripts aren't defined on this branch

lib/js/importers/AllInOneImporter.ts
```diff
@@ -0,0 +1,189 @@
...
+	let file = fsT.createWriteStream(filepath);
+	file.on('error', function(err) { console.log(err) });
+	file.write(JSON.stringify(content));
+	file.end();
+}
+
+const AllInOneImporter = () => {
+	tokensFilesConfig.bins.forEach(bin =>{
+		axios.get(tokensFilesConfig.jsonBinApiUrl + bin.id + '/latest')
+		.then(function(response) {
```
kvas-damian, line 173: `await` is easier to understand and error handling

lib/js/importers/AllInOneImporter.ts
```diff
@@ -0,0 +1,189 @@
...
+			let hexToCssVariable;
+			Object.keys(bin.files).forEach(key => {
+				switch (key) {
+					case "colorsRaw":
+						hexToCssVariable = ImportColorsRaw(bin.files[key], response);
+						break;
+					case "tokens":
+						ImportSingleTokenFile(bin.files[key], response, hexToCssVariable);
```
kvas-damian, line 181: this one is tricky. The order is dependent on the config file, on the other hand, if you call `ImportSingleTokenFile()` before `ImportColorsRaw()` you'll get an empty variable. IMO, we should mitigate this problem somehow

lib/js/importers/AllInOneImporter.ts
```diff
@@ -0,0 +1,189 @@
...
+	} catch (err) {
+		console.error(err)
+	}
+
+	return hexToCssVariable;
+};
+
+const ImportSingleTokenFile = (
+	binValues: configFileObject,
+	jsonSource: { data },
```
kvas-damian, line 85: to make method signature easier to use, we should destruct response on a higher level

lib/js/importers/configs/allInOneConfig.json
```diff
@@ -0,0 +1,48 @@
+{
+  "destinationPath": "lib/styles/settings/colors/",
+  "bins": [
+    {
+      "id": "620e4f834bf50f4b2d0044e9",
+      "name": "wnl",
+      "isTheme": false,
```
kvas-damian, line 7: is it used?

lib/js/importers/AllInOneImporter.ts
```diff
@@ -0,0 +1,189 @@
...
+	id: string;
+	label: string;
+	value: string;
+}
+
+const ImportColorsRaw = (
+	binValues: configFileObject,
+	jsonSource: { data }
+) => {
+	let hexToCssVariable: Array<Object> = [];
```
kvas-damian, line 29: I suspect this type definition is wrong. In JS, when you use string keys, then array gets converted to `{}`

kvas-damian: BTW. you put `string` values, not `Object`

lib/js/importers/AllInOneImporter.ts
```diff
@@ -0,0 +1,189 @@
...
+	try {
+		let result: Array<string> = [];
+		let resultJson = {};
+
+		jsonSource.data.colors.forEach(obj => {
+			const patternIsRawColor = /RAW\/|theme/i;
```
kvas-damian, line 35: variable name doesn't match what we check inside. The same for function name

lib/js/importers/AllInOneImporter.ts
```diff
@@ -0,0 +1,189 @@
...
+			const patternDeprecated = /deprecated|Pattern/i;
```
kvas-damian, line 36: it isn't only deprecated. For me it looks like pattern of names to be ignored

lib/js/importers/AllInOneImporter.ts
```diff
@@ -0,0 +1,189 @@
...
+				let colorName = (nameSplit[2] === undefined) ? nameSplit[1] : nameSplit[2];
+
+				obj.values.hex = obj.values.hex.replace(/FFFFFF/gi, function(){
+					return 'fff';
+				});
+
+				const colorPrefix = obj.name.match(patternTheme) ? '--' : '--raw-';
+
+				if (hexToCssVariable[obj.values.hex] === undefined) {
+					hexToCssVariable[obj.values.hex] = colorPrefix + colorName;
```
kvas-damian, line 49: we use `colorPrefix + colorName` 3 times, maybe we could create a separate variable for it?

lib/js/importers/AllInOneImporter.ts
```diff
@@ -0,0 +1,189 @@
...
+				}
+
+				result.push(colorPrefix + colorName + ': ' + obj.values.hex + ';');
+				result.push(colorPrefix + colorName + '-rgb' + ': ' + hexToRgb(obj.values.hex) + ';');
+
+				/** JSON object structure */
+				colorName = colorName.replace(/--/i, '').replace(/raw-/i, '');
```
kvas-damian, line 56: is it required? I don't see when we get `raw` or `--` into colorName

lib/js/importers/AllInOneImporter.ts
```diff
@@ -0,0 +1,189 @@
...
+				const colorNameSplitted = colorName.split('-');
+				const category = (colorNameSplitted[1] === undefined) ? 'default' : colorNameSplitted[0];
+				const resultJsonObject: IResultJsonObject = {
+					id: binValues.destination + '_' + colorName,
+					label: colorName,
+					value: obj.values.hex,
+					weight: colorNameSplitted[2] ? parseInt(colorNameSplitted[2]) : parseInt(colorNameSplitted[1])
```
kvas-damian, line 63: I believe `colorNameSplitted[2]` is not defined in current bins

lib/js/importers/configs/allInOneConfig.json
```diff
@@ -0,0 +1,48 @@
...
+        },
+        "tokens": {
+          "destination": "_tokens-bodywork.scss",
+          "destinationJson": "_tokens-bodywork.json"
+        }
+      }
+    }
+  ],
+  "jsonBinApiUrl": "https://api.jsonbin.io/b/",
+  "doubleStringsMap": {
```
kvas-damian, line 36: is it used?

lib/js/importers/AllInOneImporter.ts
```diff
@@ -0,0 +1,189 @@
...
+				};
+
+				if (resultJson[category] === undefined) {
+					resultJson[category] = [];
+				}
+				resultJson[category].push(resultJsonObject);
+				resultJson[category].sort((a,b) => (parseInt(a.weight) > parseInt(b.weight)) ? 1 : ((parseInt(b.weight) > parseInt(a.weight)) ? -1 : 0))
```
kvas-damian, line 70:
```suggestion
				resultJson[category].sort((a,b) => parseInt(a.weight) - parseInt(b.weight)
```

lib/js/importers/AllInOneImporter.ts
```diff
@@ -0,0 +1,189 @@
...
+			}
+		});
+		saveColorFile(tokensFilesConfig.destinationPath + binValues.destination, result);
+		saveColorFileJSON(tokensFilesConfig.destinationPath + binValues.destinationJson, resultJson);
+
+	} catch (err) {
+		console.error(err)
```
kvas-damian, line 77: does it makes sense to handle this error? If it happend we end up with corrupted `hexToCssVariable`, which will break other imports...

lib/js/importers/AllInOneImporter.ts
```diff
@@ -0,0 +1,189 @@
...
+	file.end();
+}
+
+const hexToRgb = (hex: string ) => {
+	if (hex.length === 4) {
+		hex = hex + hex[1] + hex[2] + hex[3];
+	}
+	let result = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex);
+	return result ? '' + parseInt(result[1], 16) + ', ' +
+		parseInt(result[2], 16) + ', ' + parseInt(result[3], 16)  : null;
```
kvas-damian, line 151: should we return `null` or throw an error here?

lib/js/importers/AllInOneImporter.ts
```diff
@@ -0,0 +1,189 @@
...
+			}
+		});
+		saveTokenFile(tokensFilesConfig.destinationPath + binValues.destination, result);
+		saveTokenFileJSON(tokensFilesConfig.destinationPath + binValues.destinationJson, resultJson);
+	} catch (err) {
+		console.error(err)
+	}
+}
+
+const saveTokenFileJSON  = (filepath: string, content: any) => {
```
kvas-damian, line 131: why `any`?

lib/js/importers/AllInOneImporter.ts
```diff
@@ -0,0 +1,189 @@
...
+	content.forEach(function(v) { file.write('\t' + v.toLowerCase() + '\n'); });
+	file.write('}\n')
+	file.end();
+}
+
+const saveColorFileJSON  = (filepath: string, content: any) => {
+	let file = fsT.createWriteStream(filepath);
+	file.on('error', function(err) { console.log(err) });
+	file.write(JSON.stringify(content));
+	file.end();
```
kvas-damian, line 167: IMO, file handling is duplicated multiple times, we should extract it to a separate method

lib/js/importers/AllInOneImporter.ts
```diff
@@ -0,0 +1,189 @@
...
+	binValues: configFileObject,
+	jsonSource: { data },
+	hexToCssVariable: Object,
+) => {
+	try {
+		let result: Array<string> = [];
+		let resultJson = {};
+
+		jsonSource.data.colors.forEach(obj => {
+			const patternDeprecated = /deprecated|Pattern|RAW|theme/i;
```
kvas-damian, line 93: it looks like `ignored` not `deprecated`

lib/js/importers/AllInOneImporter.ts
kvas-damian: Do we use `weight` after we sort values? Maybe we should remove it?

kvas-damian:
Is it ok that bodywork bin cotains only theme colors, and wnl contains all (theme + raw)? How do you plan to handle that on prod or when we have more themes?

package.json
damianmef: This line is outdated, now is `AllInOneImporter.ts`. Yes, `synchronize` is better word so I will renamed it to `SynchronizeColorsTokens.ts`

lib/js/styles/Colors/Colors.vue
damianmef: I will move styles and remove class `sb_color` , but this span is needed to set some padding.

lib/js/styles/Colors/Colors.vue
damianmef: Maybe `sb_colorsMap`?

lib/js/styles/Colors/Colors.vue
damianmef: Roger that.

lib/js/styles/Deprecated/Colors.stories.mdx
damianmef: No, I'm reverting this change.

lib/styles/modifiers/_colors.scss
damianmef: Right.

lib/styles/settings/_colors.scss
damianmef: Yes, I will revert and not touch this file more.

lib/js/styles/Colors/Colors.vue
kvas-damian: for me, it looks like a color definition, so maybe `colorDefinition`?

BTW. where did the `sb` prefix come from?

lib/js/styles/Colors/Colors.vue
damianmef: Ok, so maybe `source` in the same level like `lib`?

lib/js/styles/Colors/Colors.vue
damianmef: `sb` is from Storybook, I had no idea what prefix do we use in styles which are global in SB. This prefix is not necessary, classes `colorDefinition` `colorDefinition__id` `colorDefinition__value` looks ok @kvas-damian

lib/js/styles/Colors/Colors.vue
damianmef: So maybe `source` folder on the same level as `lib`?

lib/styles/settings/colors/_tokens.scss
damianmef: Yeah, You found the bug, there should be `--raw-gray-700-rgb` and it occures everywhere in this file. I will fix it in importer.

package.json
damianmef: As I mentioned above, now is only one script, I will update this line.

lib/js/importers/configs/allInOneConfig.json
damianmef: It was used in previous importers. I thought also it will be use in views, but views have manually defined loaded files. So I will remove it, but I suppose we can back to this option in the future.

lib/js/importers/AllInOneImporter.ts
damianmef: Thanks.

lib/js/styles/Colors/Colors.vue
damianmef: Updated In all 4 new views.

lib/js/importers/AllInOneImporter.ts
damianmef: Updated.

lib/js/importers/AllInOneImporter.ts
damianmef: Can You check now is it ok?

lib/js/importers/AllInOneImporter.ts
damianmef: Updated.

lib/js/importers/AllInOneImporter.ts
damianmef: Updated.

lib/js/importers/AllInOneImporter.ts
damianmef: Updated.

lib/js/importers/AllInOneImporter.ts
damianmef: Done.

lib/js/importers/AllInOneImporter.ts
damianmef: Updated.

source/importers/SynchronizeColorsTokens.ts
```diff
@@ -0,0 +1,191 @@
...
+
+const SynchronizeColorsTokens = () => {
+	tokensFilesConfig.bins.forEach(bin =>{
+		axios.get(tokensFilesConfig.jsonBinApiUrl + bin.id + '/latest')
+		.then(function(response) {
+			let hexToCssVariable;
+			Object.keys(bin.files).forEach(key => {
+				switch (key) {
+					case "colorsRaw":
+						hexToCssVariable = ImportColorsRaw(bin.files[key], response?.data?.colors);
```
kvas-damian, line 176: optional chaining returns `undefined`, `ImportColorsRaw()` doesn't accept undefined in type definition. Probably we should validate API response before passing it here

lib/js/importers/AllInOneImporter.ts
kvas-damian: Looks better. Now it's consistent with what's inside. IMO, it could be even better, if we name it based on it's function, like `patternColorsToProcess` and `patternColorsToIgnore`. If we do that, you already know how to use this variable and adding another pattern won't require variable rename.

lib/js/importers/AllInOneImporter.ts
damianmef: I did it. I'm sure there is smarter way, but I spend too much time to try optimise it. Maybe you have some idea. I mean casting object to another object, I was sure it will work but it didn't:
`const resultFileJsonObject = <IResultFileJsonObject> color`

lib/js/importers/AllInOneImporter.ts
damianmef: No, updated.

lib/js/importers/AllInOneImporter.ts
```diff
@@ -0,0 +1,189 @@
...
+			if (obj.name.match(patternDeprecated) === null) {
+				let tokenName = obj.name;
+				tokenName = tokenName.replace(/\//i, '-');
+
+				obj.values.hex = obj.values.hex.replace(/FFFFFF/gi, function(){
```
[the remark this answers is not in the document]
damianmef: Thanks, I missed that.

lib/js/importers/AllInOneImporter.ts
damianmef: I reorganised, from 4 function to 2, one is for line by line flies, second for json files. I think now Is acceptable.

lib/js/importers/AllInOneImporter.ts
damianmef: But it still break one bin. Until token import is depended on colours import, I think I can be break like this.

lib/js/importers/AllInOneImporter.ts
damianmef: I added fast check condition at the start of function.

lib/js/importers/AllInOneImporter.ts
damianmef: Sure, sounds clear.

lib/js/importers/AllInOneImporter.ts
damianmef: Yes, its Object anyway.

lib/js/importers/AllInOneImporter.ts
damianmef: Updated.

lib/js/importers/AllInOneImporter.ts
damianmef: I changed it, for one type of files is `content: Array<string>` for second one `content: Object`

source/importers/SynchronizeColorsTokens.ts
damianmef: Updated.

source/importers/SynchronizeColorsTokens.ts
```diff
@@ -0,0 +1,218 @@
...
+		let resultJson = {};
+
+		jsonColors.forEach(obj => {
+			const patternColorsToIgnore = /deprecated|Pattern|RAW|theme/i;
+			if (obj.name.match(patternColorsToIgnore) === null) {
+				let tokenName = obj.name;
+				tokenName = tokenName.replace(/\//i, '-');
+
+				if (obj.values.hex .split('').every(char => char === obj.values.hex [0])){
+					obj.values.hex  = `${obj.values.hex [0]}${obj.values.hex [1]}${obj.values.hex [2]}`
```
kvas-damian, line 122: these random spaces looks strange, did you consider using `prettier` on save https://bethink.atlassian.net/wiki/spaces/LMS/pages/137265159/Korzystanie+z+formatera+na+froncie ?

source/importers/SynchronizeColorsTokens.ts
```diff
@@ -0,0 +1,218 @@
...
+		console.error(err)
+	}
+
+	return hexToCssVariable;
+};
+
+const ImportSingleTokenFile = (
+	binValues : configFileObject,
+	jsonColors: Array<any>,
+	hexToCssVariable: Object,
```
kvas-damian, line 109: `Dict` interface looks more accurate here: https://github.com/bethinkpl/wnl-platform/blob/master/resources/assets/js/utils/types.utils.ts#L1

source/importers/SynchronizeColorsTokens.ts
```diff
@@ -0,0 +1,218 @@
...
+		result.push(':root {');
+
+		jsonColors.forEach(obj => {
+			const patternColorsToProcess = /RAW\/|theme/i;
+			const patternColorsToIgnore = /deprecated|Pattern/i;
+			const patternTheme = /theme/i;
+			if (obj.name.match(patternColorsToProcess) && obj.name.match(patternColorsToIgnore) === null) {
+				const nameSplit = obj.name.split('/');
+				let colorName = (nameSplit[2] === undefined) ? nameSplit[1] : nameSplit[2];
+				console.log(colorName);
```
kvas-damian, line 45:
```suggestion
```

source/importers/SynchronizeColorsTokens.ts
```diff
@@ -0,0 +1,218 @@
...
+				obj.values.hex = obj.values.hex.replace(/FFFFFF/gi, function(){
```
kvas-damian, line 46: we already found a way to handle all colors, can we apply it here?

lib/js/styles/ColorsThemes/ColorsThemes.vue
kvas-damian: instead of moving styles to separate scss file, maybe we could reuse component? By doing that we avoid HTML markup duplication

package.json
```diff
@@ -68,6 +71,7 @@
     "@fortawesome/pro-solid-svg-icons": "^5.15.3",
     "@fortawesome/vue-fontawesome": "^2.0.2",
     "storybook-addon-designs": "^6.2.1",
+    "axios": "^0.26.0",
```
kvas-damian, line 74: should be moved to `dev-dependencies`

source/importers/SynchronizeColorsTokens.ts
```diff
@@ -0,0 +1,218 @@
...
+}
+
+const ImportColorsRaw = (
+	binValues: configFileObject,
+	jsonColors: Array<any>
+) => {
+	let hexToCssVariable: Object = {};
+	try {
+		let result: Array<string> = [];
+		let temporaryColorsJson = {};
```
kvas-damian, line 33: `Dict<Array<IResultJsonObject>>`?
Similar for line below

source/importers/SynchronizeColorsTokens.ts
```diff
@@ -0,0 +1,218 @@
...
+
+				if (hexToCssVariable[obj.values.hex] === undefined) {
+					hexToCssVariable[obj.values.hex] = colorFinalName ;
+				}
+
+				result.push(colorFinalName + ': ' + obj.values.hex + ';');
+				result.push(colorFinalName + '-rgb' + ': ' + hexToRgb(obj.values.hex) + ';');
+
+				/** JSON object structure */
+				colorName = colorName.replace(/--/i, '').replace(/raw-/i, '');
```
kvas-damian, line 61: is it needed?

source/importers/SynchronizeColorsTokens.ts
```diff
@@ -0,0 +1,218 @@
...
+					label: colorName,
+					value: obj.values.hex,
+					weight: parseInt(colorNameSplitted[1])
+				};
+
+				if (temporaryColorsJson[category] === undefined) {
+					temporaryColorsJson[category] = [];
+				}
+				temporaryColorsJson[category].push(resultJsonObject);
+				temporaryColorsJson[category].sort((a,b) => a.weight - b.weight);
```
kvas-damian, line 75: you're sorting on every color push. For performance reaseons it's better to do that once, after you finish the loop

source/importers/SynchronizeColorsTokens.ts
```diff
@@ -0,0 +1,218 @@
...
+		});
+		arrayToFile(tokensFilesConfig.destinationPath + binValues.destination, result);
+		jsonToFile(tokensFilesConfig.destinationPath + binValues.destinationJson, resultJson);
+	} catch (err) {
+		console.error(err)
+	}
+}
+
+
+const hexToRgb = (hex: string ) => {
```
kvas-damian, line 153: this method returns `null` now. Are you sure, you handle that correctly in all places?

source/importers/SynchronizeColorsTokens.ts
```diff
@@ -0,0 +1,218 @@
...
+	file.on('error', function(err) { console.log(err) });
+	file.write(JSON.stringify(content));
+	file.end();
+}
+
+const SynchronizeSingleBin = async (bin) => {
+	try {
+		const response = await axios.get(tokensFilesConfig.jsonBinApiUrl + bin.id + '/latest');
+		let hexToCssVariable;
+		Object.keys(bin.files).forEach(() => {
```
kvas-damian, line 195: do we need this `forEach()`? Aren't we performing the same operations twice now?

package.json
```diff
@@ -23,14 +23,16 @@
     "storybook": "start-storybook -p 6006",
     "storybook:build": "build-storybook -c .storybook -o public/storybook",
     "test": "jest",
-    "ts:check": "tsc --noEmit"
+    "ts:check": "tsc --noEmit",
+    "import": "npx ts-node source/importers/SynchronizeColorsTokens.ts"
```
kvas-damian, line 27:
```suggestion
    "synchronizeColors": "npx ts-node source/importers/SynchronizeColorsTokens.ts"
```

kvas-damian:
IMO, error handling is still broken in this code. When I run this script, I expect it to finish when all is ok, or **fail without saving corrupted data to file**. Currently, I see opposite approach, log/silence error and proceed - it requires end users to manually verify the output on every execution.

1. `ImportSingleTokenFile()` is dependant on `hexToCssVariable` which is calculated in `ImportColorsRaw()`. In `ImportColorsRaw()` we have a big `try/catch` block, which logs an error and continue (returns `hexToCssVariable` with missing data). Then we pass this corrupted variable to `ImportSingleTokenFile()`.
2. `hexToRgb()` can return `null` which is saved to output css file. It isn't correct CSS, so it should not be put there.

source/importers/SynchronizeColorsTokens.ts
damianmef: True, I'm removing this loop.

package.json
damianmef: Updated.

source/importers/SynchronizeColorsTokens.ts
damianmef: Updated.

source/importers/SynchronizeColorsTokens.ts
damianmef: Updated.

source/importers/SynchronizeColorsTokens.ts
damianmef: Yes, I set return null always and it works. What you mean all places? I has one call only here:
`result.push(colorFinalName + '-rgb' + ': ' + hexToRgb(obj.values.hex) + ';');`

damianmef: Wait, it occures problems however.

damianmef: I fixed it. If rgb is not calculated, this rgba color is not pushed to final list.

source/importers/SynchronizeColorsTokens.ts
damianmef: It's no more needed.

source/importers/SynchronizeColorsTokens.ts
damianmef: Thanks. I used it. I would like to make interfaces and function resuable, but not now. In next usage in DS in similar script.

kvas-damian:
> Inside ImportColorsRaw and ImportSingleTokenFile I added check if any color was added to the list. If not, files are not updated.

in `ImportColorsRaw()`, Iteration over an array of colors is inside of a big try catch block. If something fails during the iteration you won't save data into a file, but you return partial data stored in `hexToCssVariable`. Is it ok?

> hexToRgb() if this function can't calculate rgb, the rgb row is not added to color list.

yep, it isn't added to the file, but you don't have a guarantee that this `raw` color isn't used in tokens. In worst scenario, tokens will be pointing to not defined RAW RGB variables.

Summarizing: If I write this code, I'd remove all `try/catch` and let code stop execution on every unexpected scenario.

source/importers/SynchronizeColorsTokens.ts
damianmef: Updated.

source/importers/SynchronizeColorsTokens.ts
damianmef: Updated.

lib/js/styles/Colors/Colors.vue
kvas-damian: `source` suggests some source code of the package. Maybe we could name it `scripts` or `tools`?

lib/js/importers/AllInOneImporter.ts
kvas-damian: `weight` is not used now. Maybe we should remove it?

source/importers/SynchronizeColorsTokens.ts
```diff
@@ -0,0 +1,243 @@
...
+						tokenName +
+						': ' +
+						'rgba(var(' +
+						hexToCssVariable[obj.values.hex] +
+						'-rgb), ' +
+						obj.values.alpha +
+						');',
+				);
+			} else {
+				result.push('$' + tokenName + ': ' + hexToCssVariable[obj.values.hex] + ';');
```
kvas-damian, line 139: don't we need `var()` keyword here?

source/importers/SynchronizeColorsTokens.ts
```diff
@@ -0,0 +1,243 @@
...
+	jsonColors.forEach((obj) => {
+		const patternColorsToIgnore = /deprecated|Pattern|RAW|theme/i;
+		if (obj.name.match(patternColorsToIgnore) === null) {
+			let tokenName = obj.name;
+			tokenName = tokenName.replace(/\//i, '-');
+			obj.values.hex = makeHexShortcut(obj.values.hex);
+
+			if (obj.values.alpha !== 1) {
+				result.push(
+					'$' +
```
kvas-damian, line 129: I'm wondering if it's ok to define tokens as SASS variables. For it, it doesn't really matter now and we can migrate that in future, but for UX it can be important.

Let's consider alternative approach: define css variable for token and then define SASS variable to that Css variable
eg. 
```
--default-border: var(--raw-black);
$default-border: var(--default-border);
```

In this scenario, UX will be able to see token name used in CSS when using devtools inspect.

What do you think about it?

lib/styles/storybook.scss
```diff
@@ -12,7 +13,7 @@ body {
 	-webkit-text-size-adjust: 100%;
 	-webkit-font-smoothing: antialiased;
 	font-family: 'Lato', sans-serif;
-
+	//background: $default-text;
```
kvas-damian, line 16:
```suggestion
```

lib/styles/settings/colors/_tokens-bodywork-variables.scss
```diff
@@ -0,0 +1,41 @@
+:root {
```
kvas-damian, line 1: `.theme-bodywork`?

lib/js/atoms/TabItem/TabItem.vue
```diff
@@ -9,8 +9,7 @@
 	</div>
 </template>
 
-<style scoped lang="scss">
-@import '../../../styles/settings/colors';
+1
```
kvas-damian, line 12: something bad happened here

lib/js/styles/ColorsList.vue
```diff
@@ -0,0 +1,61 @@
...
+					<h3 :id="'category-' + colorCategory">{{ colorCategory }}</h3>
+					<div v-for="color in colors" :key="color.id" class="colorDefinition">
+						<div class="colorDefinition__id">{{ color.label }}</div>
+						<div class="colorDefinition__value">{{ color.value }}</div>
+						<div class="colorDefinition__color">
+							<span
+								v-if="isHexOrRGBA(color.value)"
+								:style="{ background: color.value }"
+							></span>
+							<span v-else :style="{ background: 'var(' + color.value + ')' }"></span>
```
kvas-damian, line 31: when we fix exported variabled, then this conditiona handling won't be necessary

lib/js/styles/ColorsTokensBodywork/ColorsTokensBodywork.stories.mdx
```diff
@@ -0,0 +1,30 @@
...
+	}}
+	argTypes={{
+		tokensBodywork: Object,
+	}}
+>
+	{(args) => {
+		return {
+			components: { ColorsTokensBodywork },
+			template:
+				'<colors-tokens-bodywork :tokensBodywork="tokensBodywork"></colors-tokens-bodywork>',
```
kvas-damian, line 22: Instead of using separate component for theme you can add class to existing component:
`template: '<ColorsList class="theme-bodywork" :colorsLists="tokensBodywork" />',`

lib/styles/components/_colors.scss
```diff
@@ -0,0 +1,53 @@
+.colorsList {
```
kvas-damian, line 1: Now, it can be moved to `ColorsList.vue`

lib/styles/settings/colors/_tokens-bodywork-variables.scss
```diff
@@ -0,0 +1,41 @@
...
+	--primary-text: var(--theme-700);
```
kvas-damian, line 2: Don't we need similar file for WNL tokens? According to my knowledge, they're also theme specific

lib/js/styles/ColorsTokensBodywork/ColorsTokensBodywork.stories.mdx
damianmef: I made in in different way, because of one view where we display many themes: wnl, bodywork - there class need be added on the singiel list level.

lib/styles/settings/colors/_tokens-bodywork-variables.scss
damianmef: All wnl variables are in default file `_tokens-variables.scss`. We don't have wnl tokens, we have default and bodywork tokens. I think it's not a problem now.

lib/js/styles/ColorsList.vue
damianmef: I would not want change json's structures:
```    {
      "id": "_raw-wnl.scss_theme-50",
      "label": "theme-50",
      "value": "#ECF2F5"
    },```
  Of course it's possible to add there additional fields variable name, but I think is not worth now do this for removing one condition in the view.

File: lib/js/atoms/TabItem/TabItem.vue (hunk 47 above)
Comment (damianmef) [reply lib/js/atoms/TabItem/TabItem.vue]: Fixed.

File: lib/styles/storybook.scss (hunk 45 above)
Comment (damianmef) [reply lib/styles/storybook.scss]: Fixed.

File: lib/styles/settings/colors/_tokens-bodywork-variables.scss (hunk 46 above)
Comment (damianmef) [reply lib/styles/settings/colors/_tokens-bodywork-variables.scss]: Fixed.

File: source/importers/SynchronizeColorsTokens.ts (hunk 44 above)
Comment (damianmef) [reply source/importers/SynchronizeColorsTokens.ts]: It's done.

File: source/importers/SynchronizeColorsTokens.ts (hunk 43 above)
Comment (damianmef) [reply source/importers/SynchronizeColorsTokens.ts]: Fixed.

File: lib/js/styles/ColorsThemes/ColorsThemes.vue (hunk 7 above)
Comment (damianmef) [reply lib/js/styles/ColorsThemes/ColorsThemes.vue]: It's done.

File: lib/js/importers/AllInOneImporter.ts (hunk 22 above)
Comment (damianmef) [reply lib/js/importers/AllInOneImporter.ts]: Removed.

File: lib/js/styles/Colors/Colors.vue (hunk 5 above)
Comment (damianmef) [reply lib/js/styles/Colors/Colors.vue]: `tools` for me sounds good. I will change it.

File: lib/js/styles/ColorsList.vue (hunk 48 above)
Comment (kvas-damian) [reply lib/js/styles/ColorsList.vue:31]: sorry, I forgot it's json

Comment (kvas-damian) [reply lib/js/styles/ColorsList.vue:31]: On the other hand, it's inconsistency in displayed value

Zaznaczenie_717

For some colors we use variable name which cannot be used directly in code and for others we have `rgba()` signature. It can be confusing, why sometimes it should be directly copied and sometimes modified.

File: lib/js/styles/ColorsList.vue [hunk 52]
```diff
@@ -0,0 +1,133 @@
+<template>
+	<div>
+		<div v-for="colorList in colors" :key="colorList.title" class="colorDefinition">
+			<div class="colorsList" :class="{ [colorList.class]: true }">
```
kvas-damian, line 4: when class is not defined it generates `undefined`

lib/js/styles/ColorsList.vue
```diff
@@ -0,0 +1,133 @@
...
+	div {
+		min-width: 200px;
+		padding: 10px;
+	}
+
+	&__color {
+		flex: 3 1 auto;
+		width: 200px;
+
+		span {
```
kvas-damian, line 75: try to avoid styling based on tag name selector

tools/importers/SynchronizeColorsTokens.ts
```diff
@@ -0,0 +1,304 @@
...
+const ImportSingleTokenFile = (
+	binValues: configFileObject,
+	jsonColors: Array<any>,
+	hexToCssVariable: Dict<string>,
+	themeName: string,
+	isTheme: boolean,
+) => {
+	let result: Array<string> = [];
+	let resultJsonTokens = {};
+	let resultJsonVariables: Array<string> = [];
```
kvas-damian, line 132: should it be `resultJsonVariables`?

lib/js/styles/ColorsList.vue
damianmef: Fixed, now all can be copied to dev tools as value.

lib/js/styles/ColorsList.vue
damianmef: Fixed.

lib/js/styles/ColorsList.vue
damianmef: Fixed all html tags in scss.

tools/importers/SynchronizeColorsTokens.ts
damianmef: No, it's not connected to json, fixed.

kvas-damian (approved):

kvas-damian (approved):

PR #99, opened 2022-03-09
create typings when building production

tweetgeek:
From now, we also build typings for typescript during production building, thanks to which the IDE is very good at prompting and understanding DS components. And it does not glow red anymore :)

Then:
Zrzut ekranu 2022-03-9 o 13 03 07

Now
Zrzut ekranu 2022-03-9 o 13 07 58


.gitignore
```diff
@@ -6,4 +6,3 @@ yarn-error.log
 .stylelintcache
 .npmrc
 public
-dist
```
rogatty, line 9: Are you sure this is needed? Can it be built during the release? See https://github.com/bethinkpl/design-system/blob/master/.github/workflows/build.yml

kvas-damian, line 9: we're building `dist` after merge to master, to avoid conflicts with other branches. Can we revert this change?

vue.config.js
```diff
@@ -1,8 +1,28 @@
 module.exports = {
+	parallel: process.env.NODE_ENV !== 'production',
```
kvas-damian, line 2: can you add a comment why we disable it?

tweetgeek: added :)

dist/demo.html
```diff
@@ -1,8 +0,0 @@
-<meta charset="utf-8">
```
rogatty, line 1: I'd revert `dist/` to the original state, just to avoid the noise, because I don't think it matters in the end.

kvas-damian (approved):

dist/demo.html
kvas-damian: [`1042680` (#99)](https://github.com/bethinkpl/design-system/pull/99/commits/1042680e5eff8e919f7771ec1abe21bff706a7a6)

PR #103, opened 2022-03-22
IT-4331 | CardExpandable

kvas-damian:


lib/js/components/Cards/CardExpandable.vue
```diff
@@ -0,0 +1,124 @@
...
+	&__expander {
+		align-items: center;
+		background-color: $neutral-background;
+		display: flex;
+		height: 44px;
+		justify-content: center;
+		padding: $space-xxs $space-s;
+
+		&:hover {
+			cursor: pointer;
```
pioziol123, line 37: `cursor` will show up only on `hover` so imo it can be moved up into the `$__expander` class

kvas-damian: [`a96a7c6` (#103)](https://github.com/bethinkpl/design-system/pull/103/commits/a96a7c61c1ab3ab8568c970e5bd45ca9a121b546)

lib/js/components/Cards/CardExpandable.vue
```diff
@@ -0,0 +1,124 @@
...
+			}
+
+			return this.isExpandedInternal ? 'Zwiń' : 'Rozwiń';
+		},
+	},
+	watch: {
+		isExpanded: {
+			handler(isExpanded) {
+				if (isExpanded !== this.isExpandedInternal) {
+					this.isExpandedInternal = isExpanded;
```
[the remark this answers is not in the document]
kvas-damian: https://vuejs.org/api/options-lifecycle.html#updated
> Called after the component has updated its DOM tree due to a reactive state change.

We don't use this prop in template, so there is no update in DOM tree => `updated` won't be called when value is changed

BTW, I don't see the benefit of having this logic in `updated` hook. Can you give me more context?

pioziol123 (approved):

PR #107, opened 2022-03-29
Colors migration

damianmef:


lib/js/components/Card/Card.vue
```diff
@@ -14,9 +14,10 @@
 @import '../../../styles/settings/spacings';
 @import '../../../styles/settings/shadows';
 @import '../../../styles/settings/radiuses';
+@import '../../../styles/settings/colors/tokens';
 
 .card {
-	background-color: $color-total-white;
+	background-color: $color-default-background;
```
kvas-damian, line 20: `@import '../../../styles/settings/colors';` is not used now. Please delete it here and in all other components

lib/js/components/buttons/Button/Button.consts.ts
```diff
@@ -10,17 +10,21 @@ export const BUTTON_TYPES = {
 	TEXT: 'text',
 } as const;
 
-export const BUTTON_COLORS = {
+export const BUTTON_ICON_COLORS = {
```
kvas-damian, line 13: IMO, it's easier to maintain when you join additional colors with `BUTTON_COLORS` instead of defining them twice

lib/js/components/buttons/IconButton/IconButton.consts.ts
```diff
@@ -16,11 +15,9 @@ export const ICON_BUTTON_TYPES = {
 
 export const ICON_BUTTON_COLOR_SCHEMES = {
 	ALL_IN_COLOR: 'all-in-color',
-	MINOR_ICON: 'minor-icon',
-	MINOR_LABEL: 'minor-label',
+	NEUTRAL_LABEL: 'neutral-label',
 } as const;
 
 export const ICON_BUTTON_COLORS = {
```
kvas-damian, line 21: `ICON_BUTTON_COLORS` vs `BUTTON_ICON_COLORS` what's the difference? Do we need to have 2 variables?

lib/js/components/buttons/IconButton/IconButton.vue
```diff
@@ -31,61 +31,68 @@
...
 
+@mixin iconButtonColor($color, $icon, $border, $background, $hovered, $focused, $disabled) {
 	color: $color;
+	border-color: $border;
+	background-color: transparent;
 
 	&:hover,
 	&:active {
-		color: $hover-color;
+		//background-color: $hovered;
```
kvas-damian, line 55: why these are commented out? Don't we need them to get correct colors when user interacts with this component?

lib/styles/settings/_colors.scss
```diff
@@ -1,89 +1,80 @@
...
-
-$color-primary-rgb: var(--color-primary-rgb);
-$color-primary-supporting-rgb: var(--color-primary-supporting-rgb);
-$color-primary-opacity25: rgba(var(--color-primary-rgb), 0.25);
-$color-secondary-rgb: var(--color-secondary-rgb);
-$color-secondary-supporting-rgb: var(--color-secondary-supporting-rgb);
-$color-total-white-opacity50: rgba(255, 255, 255, 0.5);
+@import '../../styles/settings/colors/raw';
+
+$color-firefly-black: var(--raw-black);
```
kvas-damian, line 3: Can you share your plan for these variables?

lib/styles/components/_buttons.scss
```diff
@@ -0,0 +1,246 @@
...
+@mixin buttonColors(
+	$color: '',
+	$background: '',
+	$hover: '',
+	$focus: '',
+	$disabled: '',
+	$ripple: '',
+	$border: ''
+) {
+	@if $color != '' {
```
kvas-damian, line 19: should this color be optional? I have a feeling that when it's missing, then we have a bug

lib/styles/components/_buttons.scss
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	@if $background != '' {
+		background-color: $background;
+	} @else {
+		background-color: transparent;
+	}
+
+	@if $hover != '' {
+		@if $border != '' or $background != '' {
```
kvas-damian, line 30: `if` should go inside selector block

lib/styles/components/_buttons.scss
```diff
@@ -0,0 +1,246 @@
...
+		&:focus,
+		&.-focused {
+			background-color: $focus;
+		}
+	}
+
+	@if $disabled != '' {
+		&:disabled,
+		&.-disabled {
+			background-color: $disabled;
```
kvas-damian, line 57: disabled outlined doesn't work now

lib/styles/components/_buttons.scss
```diff
@@ -0,0 +1,246 @@
...
+	}
+}
+
+$button-small-min-height: 28px;
+$button-medium-min-height: 32px;
+$button-large-min-height: 48px;
+
+$button-hover-alpha: 0.12;
+$button-focus-alpha: 0.24;
+$button-disabled-alpha: 0.6;
```
kvas-damian, line 78: these variables are not used now

lib/styles/settings/_buttons.scss
```diff
@@ -48,3 +49,162 @@ $regular-colors: (
...
+$only-icon-button-colors: (
+	'neutral-weak': (
+		'color': $color-default-text-inverted,
+		'icon': $color-default-icon-inverted,
+		'background': $color-neutral-background-strong,
+		'hovered': $color-neutral-background-strong-hovered,
+		'focused': $color-neutral-background-strong-focused,
+		'disabled': $color-neutral-background-strong-disabled,
+		'ripple': $color-default-ripple-inverted,
+		'inverted': (
```
kvas-damian, line 62: UX uses `inverted` work with different meaning. IMO, we should think of a better name here

lib/styles/components/_buttons.scss
kvas-damian: disabled text is also strange

lib/styles/modifiers/_colors.scss
```diff
@@ -2,10 +2,6 @@
 	color: $color-cadet-gray;
 }
 
-.-catalinaBlue {
```
kvas-damian, line 5: Are you sure it's backwards compatible change?

lib/js/components/buttons/IconButton/IconButton.vue
```diff
@@ -282,7 +282,7 @@ export default {
 			throw new Error(`Color: ${this.color} is supported only in type: "icon-only"`);
 		},
 		isButtonColor(): boolean {
```
kvas-damian, line 284: From what I see, all available colors for IconButton are defined for Button. IMO, we can remove this function.
The same comment applies to setting color directly in `style` attribute in this component

kvas-damian:
Linter is failing

lib/js/components/Badges/BadgeScore/BadgeScore.spec.ts
```diff
@@ -64,28 +66,31 @@ describe('BadgeScore', () => {
 		expect(component.find('.-success').exists()).toBe(false);
 		expect(component.find('.-warning').exists()).toBe(false);
 		expect(component.find('.-danger').exists()).toBe(true);
```
kvas-damian, line 68: there is no `danger` anymore. Does it still work?

PR #115, opened 2022-04-10
Colors migration - 2a - Small components

damianmef:


lib/js/components/Survey-Questions/SurveyQuestionScale/SurveyQuestionScale.stories.mdx
```diff
@@ -80,140 +86,140 @@ import { SURVEY_QUESTION_STATES } from '../SurveyQuestion.consts';
...
-				},
-			},
-			template:
-				'<div style="max-width: 600px"><survey-question-scale :title="title" :scale-options="scaleOptions" :elaboration-value="elaborationData" :elaborationLabel="elaborationLabel" :placeholder="placeholder" :selected-value="selectedValue" :state="state" @select-change="selectedValue = $event" @elaboration-change="elaborationData = $event">' +
-				'<div v-if="explanation" slot="explanation" v-html="explanation" />' +
-				'</survey-question-scale></div>',
-		};
-	}}
-</Story>
+        <Story
```
damianmef, line 89: @kvas-damian can You check in your IDE is it correct?

lib/js/components/Badges/BadgeScore/BadgeScore.spec.ts
```diff
@@ -43,49 +43,54 @@ describe('BadgeScore', () => {
...
+		const component = createComponent({ color: BADGE_SCORE_COLORS.NEUTRAL });
 
 		expect(component.find('.-success').exists()).toBe(false);
 		expect(component.find('.-warning').exists()).toBe(false);
-		expect(component.find('.-danger').exists()).toBe(false);
-		expect(component.find('.-minor').exists()).toBe(true);
+		expect(component.find('.-fail').exists()).toBe(false);
+		expect(component.find('.-neutral').exists()).toBe(true);
 		expect(component.find('.-inverted').exists()).toBe(false);
+		expect(component.find('.-primary').exists()).toBe(false);
```
kvas-damian, line 93: where does `-primary` class come from?

lib/styles/components/_badge-score.scss
```diff
@@ -0,0 +1,29 @@
+@import '../../styles/settings/colors/tokens';
+
+$badge-score-colors: (
```
kvas-damian, line 3: does it needs to be placed in a separate file?

lib/js/components/Badges/BadgeScore/BadgeScore.vue
```diff
@@ -1,60 +1,71 @@
...
 @import '../../../../styles/settings/colors';
+@import '../../../../styles/components/badge-score';
+@import '../../../../styles/settings/colors/tokens';
 
 $badge-score-width: 74px;
 $small-badge-score-width: 48px;
 $x-small-badge-score-width: 36px;
 
-.a-badgeScore {
+@mixin setBadgeScoreColor($self, $border: '', $color: '', $color-additional: '') {
```
kvas-damian, line 29: should these arguments be optional?

lib/js/components/Badges/BadgeScore/BadgeScore.vue
```diff
@@ -1,60 +1,71 @@
...
+				$self,
+				map-get($color-map, 'border'),
+				map-get($color-map, 'color'),
+				map-get($color-map, 'color-additional')
+			);
+		}
+	}
+
 	border-radius: 4px;
 	border: 2px solid currentColor;
```
kvas-damian, line 63: it isn't `currentColor` anymore

lib/js/components/Modal/Modal.vue
```diff
@@ -73,6 +74,7 @@
 		z-index: 1;
 
 		&__close:only-child {
+			color: $color-neutral-icon-weak;
```
kvas-damian, line 77: are you sure this color should be placed in this attribute?

lib/js/components/Modal/Modal.vue
```diff
@@ -85,6 +87,7 @@
 	&__content {
 		overflow: auto;
 		padding: $space-s;
+		color: $color-default-text;
```
kvas-damian, line 90: do we have to everride default value defined it https://github.com/bethinkpl/design-system/blob/57a49b3fcc4417303d554b74977bd927d11445a6/lib/styles/design-system.scss#L40 or is it enough to fix this line?

lib/js/components/NumberInCircle/NumberInCircle.vue
```diff
@@ -36,32 +37,44 @@
...
-		color: $color-primary;
-	}
-
-	&.-is2 {
-		color: $color-salad-green;
+	&.-is1,
+	&.-is6,
+	&.-primary {
+		color: $color-primary-text;
+		border-color: $color-primary-border;
```
kvas-damian, line 44: we should get rid of `currentColor` in default state now

lib/js/components/Statuses/BlockadeStatus/BlockadeStatus.vue
```diff
@@ -9,20 +9,27 @@
 @import '../../../../styles/settings/colors';
```
kvas-damian, line 9: why don't you remove this import?

lib/js/components/Survey-Questions/SurveyQuestionOpenEnded/SurveyQuestionOpenEnded.vue
```diff
@@ -46,8 +46,11 @@
 @import '../../../../styles/settings/buttons';
 @import '../../../../styles/settings/spacings';
 @import '../../../../styles/settings/typography';
+@import '../../../../styles/settings/colors/tokens';
 
 .surveyQuestionOpenEnded {
+	color: $color-default-text;
```
kvas-damian, line 52: this color should be here by default. Why do we need to override it?

kvas-damian: Please verify this problem in all places

lib/js/components/Buttons/IconButton/IconButton.consts.ts
```diff
@@ -21,6 +20,12 @@ export const ICON_BUTTON_COLOR_SCHEMES = {
 } as const;
 
 export const ICON_BUTTON_COLORS = {
-	...COLORS,
-	...BUTTON_COLORS,
+	PRIMARY: 'primary',
+	NEUTRAL: 'neutral',
```
kvas-damian, line 24: after this change story for `IconButton` won't work correctly. I see multiple usages of `BUTTON_COLORS.MINOR` there

lib/js/components/Survey-Questions/SurveyQuestionScale/SurveyQuestionScale.stories.mdx
damianmef: The problem is in the plugin `MDX`. When I disable it, everything is ok. So I'll align the text only.

lib/js/components/Badges/BadgeScore/BadgeScore.spec.ts
damianmef: Good catch, it doesn't exists.

lib/styles/components/_badge-score.scss
damianmef: @kvas-damian I added it only for one reason, other "structures" with tokens have separated files f.e. `Button` , `IconButton`, `SurveyToggle` (in this PR it is not visible). I think it's good practice, main files are smaller.

lib/js/components/Badges/BadgeScore/BadgeScore.vue
damianmef: No, it's redundant.

lib/js/components/Badges/BadgeScore/BadgeScore.vue
damianmef: I will remove it.

lib/js/components/Modal/Modal.vue
damianmef: Why not? This selector is for `svg`. You suggest:
```suggestion
		&__close {
			color: $color-neutral-icon-weak;

			&:only-child {
				margin-left: auto;
			}
		}
```

lib/js/components/Modal/Modal.vue
damianmef: If this line
```
color: var($color-default-text); 
```
is not overridden in platform by anything else, it's ok. But I'm aware some parent/rule can set some different color.

lib/js/components/Modal/Modal.vue
kvas-damian: yes, if you don't separate these style blocks and put content in `#header` slot, then the close icon would have different color

lib/js/components/NumberInCircle/NumberInCircle.vue
damianmef: Good catch, I'm removing all usages. There was only problem in `Modal.vue` - it helps to catch usage old colors.

lib/js/components/Statuses/BlockadeStatus/BlockadeStatus.vue
damianmef: Good catch, I'm removing all usages. There was only problem in `Modal.vue` - it helps to catch usage old colors.

lib/js/components/Survey-Questions/SurveyQuestionOpenEnded/SurveyQuestionOpenEnded.vue
damianmef: @kvas-damian I know the reason. Rules from `design-system.scss` are not applied in SB. Rules from `storybook.scss` are applied. It's weird because theses files are imported in the same file `preview.js`:
```
import 'design-system/lib/styles/design-system.scss';
import 'design-system/lib/styles/storybook.scss';
```

lib/js/components/Buttons/IconButton/IconButton.consts.ts
damianmef: I'm moving Your comment to IconButton PR.

lib/styles/components/_badge-score.scss
kvas-damian: IMO, when such structure is used only in one file/component such separation makes it harder to fully understand the code - you have to switch between tabs.
Having that said, I have to proofs to confirm that. It's just my feeling, so I won't force you to change that

lib/js/components/Modal/Modal.vue
kvas-damian: So what's your suggestion? Do you plan to override default text color in every component?

lib/js/components/Survey-Questions/SurveyQuestionOpenEnded/SurveyQuestionOpenEnded.vue
kvas-damian: yes, I mentioned that in https://github.com/bethinkpl/design-system/pull/115#discussion_r848372348 ;)

.storybook/preview.js
```diff
@@ -1,6 +1,6 @@
 import 'design-system/lib/styles/_normalize.scss';
 import 'design-system/lib/styles/design-system.scss';
-import 'design-system/lib/styles/storybook.scss';
+// import 'design-system/lib/styles/storybook.scss';
```
kvas-damian, line 3: why are we commenting it out?

lib/js/components/SurveyQuestions/SurveyQuestionOpenEnded/SurveyQuestionOpenEnded.vue
```diff
@@ -79,13 +80,13 @@
 </style>
 
 <script lang="ts">
-import DsCard from '../../Cards/Card';
+import DsCard from '../../Cards/Card/Card';
```
kvas-damian, line 83: we should be consisten and use `index.ts`. It means, we should revert this line

lib/js/components/Modal/Modal.vue
```diff
@@ -66,14 +66,18 @@
 
 	&__header {
 		align-items: center;
-		box-shadow: 0 $space-xxxxs $space-xxxxs $color-total-white;
+		box-shadow: 0 $space-xxxxs $space-xxxxs var(--raw-white);
```
kvas-damian, line 69: did you confirm with UX that we can use raw color here?

lib/js/components/SurveyQuestions/SurveyQuestionScale/SurveyQuestionScale.vue
```diff
@@ -82,6 +82,7 @@
 @import '../../../../styles/settings/shadows';
 @import '../../../../styles/settings/spacings';
 @import '../../../../styles/settings/typography';
+@import '../../../../styles/settings/colors/tokens';
```
kvas-damian, line 85: this file still imports old colors

lib/js/components/SurveyQuestions/SurveyQuestionTextarea.vue
```diff
@@ -14,21 +14,22 @@
 @import '../../../styles/settings/radiuses';
 @import '../../../styles/settings/spacings';
 @import '../../../styles/settings/typography';
+@import '../../../styles/settings/colors/tokens';
```
kvas-damian, line 17: this file still imports old colors

.storybook/preview.js
damianmef: Typo, reverted.

lib/js/components/SurveyQuestions/SurveyQuestionOpenEnded/SurveyQuestionOpenEnded.vue
damianmef: True, fixed.

lib/js/components/Modal/Modal.vue
damianmef: No. Modals have still `#FFFFFF` color: https://www.figma.com/file/izQdYyiBR1GQgFkaOIfIJI/LMS---DS---Components?node-id=899%3A23641
So usign raw white here I think it's ok. Probably modals will have tokens in the future, but now I don't see them in the figma, so raw is the best option.

lib/js/components/SurveyQuestions/SurveyQuestionScale/SurveyQuestionScale.vue
damianmef: Thanks, I missed it, fixed.

lib/js/components/SurveyQuestions/SurveyQuestionTextarea.vue
damianmef: Thanks, I missed it, fixed.

lib/styles/components/_badge-score.scss
damianmef: I'm also not forcing my change. It's not logic change, so reverting it is a moment. I'm also not sure it's more clear, let it stay as it is and if it will give us any problems with code reading, we can back to previous setting.

kvas-damian (approved):

PR #116, opened 2022-04-10
Colors migration - 2b - Migrate Button

damianmef:


lib/js/components/Buttons/IconButton/IconButton.vue
```diff
@@ -52,7 +52,7 @@
 
 @mixin iconButtonColor($color, $hover-color: '') {
 	@if $hover-color == '' {
-		$hover-color: mix($color, $color-firefly-black, (1 - $button-hover-alpha) * 100%);
+		//$hover-color: mix($color, $color-firefly-black, (1 - $button-hover-alpha) * 100%);
```
damianmef, line 55: It will be replaced in next PR

lib/styles/modifiers/_colors.scss
```diff
@@ -2,10 +2,6 @@
 	color: $color-cadet-gray;
 }
 
-.-catalinaBlue {
```
damianmef, line 5: I will revert it.

lib/styles/settings/_buttons.scss
```diff
@@ -16,6 +17,402 @@ $button-ripple-alpha: 0.16;
...
+			'background': $color-primary-background-ghost,
+			'background-hovered': $color-primary-background-ghost-hovered,
+			'background-focused': $color-primary-background-ghost-focused,
+			'background-disabled': $color-primary-background-ghost-disabled,
+			'ripple': $color-primary-ripple,
+			'disabled': (
+				'color': $color-primary-text-disabled,
+				'border': $color-primary-border-disabled,
+				'icon': $color-primary-icon-disabled,
+				'background': 'transparent',
```
kvas-damian, line 205: why do we sometimes have `''` and sometimes `'transparent'` as a `background-color`?

lib/styles/components/_buttons.scss
```diff
@@ -3,48 +3,183 @@
 @import '../settings/spacings';
 @import '../settings/radiuses';
 @import '../settings/typography';
+@import '../settings/buttons';
 
-$default-lighter-shadow: 0 4px 12px rgba($color-firefly-black, 0.35);
+$default-lighter-shadow: 0 4px 12px rgba($color-firefly-black-rgb, 0.35);
```
kvas-damian, line 8: firefly? Isn't it banned now?

lib/styles/components/_buttons.scss
```diff
@@ -3,48 +3,183 @@
...
+	$background: '',
+	$background-hover: '',
+	$background-focus: '',
+	$background-disabled: ''
+) {
+	@if $background != '' {
+		background-color: $background;
+	}
+
+	@if $background == '' {
```
kvas-damian, line 48: https://bethinkteam.slack.com/archives/CFBLF5RGA/p1649769822332039 we should use correct tokens instead of `transparent`

lib/styles/components/_buttons.scss
```diff
@@ -3,48 +3,183 @@
...
+			color: $color-disabled;
+		}
 	}
 }
 
-@mixin textColor($color, $hover-color: '', $focus-color: '', $disabled-color: '') {
-	@if $hover-color == '' {
-		$hover-color: mix($color, $color-firefly-black, (1 - $button-hover-alpha) * 100%);
+@mixin setButtonBackground(
+	$background: '',
```
kvas-damian, line 39: making all fields optional doesn't tell devs which are required and which can be ommited. Now, if I ommit wrong fields, we'll end up with bug. Do you think all should be optional?

lib/styles/components/_buttons.scss
```diff
@@ -3,48 +3,183 @@
...
+) {
+	@if $ripple != '' {
+		.ripple {
+			background-color: $ripple !important;
+		}
+	}
+
+	@if $ripple == '' {
+		.ripple {
+			background-color: transparent !important;
```
kvas-damian, line 129: maybe we could use `display:none` ?

lib/styles/components/_buttons.scss
```diff
@@ -3,48 +3,183 @@
...
+		}
+	}
+
+	@if $border != '' {
+		border: 1px solid $border;
+	}
+
+	@if $icon != '' {
+		.a-iconButton, // this rule here avoid repeating similar loop in IconButton
```
kvas-damian, line 138: I'm not sure about mixing these components. Personally, I prefer the same loop two times over having styles mixed in different component file

lib/styles/components/_buttons.scss
```diff
@@ -56,50 +191,54 @@ $button-hover-alpha: 0.12;
...
+		map-get($button-colors, 'primary', 'filled', 'disabled', 'icon'),
+		map-get($button-colors, 'primary', 'filled', 'disabled', 'background')
+	);
+	@include setButtonAdditions(
+		map-get($button-colors, 'primary', 'filled', 'ripple'),
+		'',
+		map-get($button-colors, 'primary', 'filled', 'icon')
+	);
+
+	@each $color-name, $color-map in $icon-button-colors {
```
kvas-damian, line 216: Now, I see that `primary` is defined two times. Can we exclude it from the loop?

lib/styles/components/_buttons.scss
```diff
@@ -3,48 +3,183 @@
...
 @import '../settings/typography';
+@import '../settings/buttons';
 
-$default-lighter-shadow: 0 4px 12px rgba($color-firefly-black, 0.35);
+$default-lighter-shadow: 0 4px 12px rgba($color-firefly-black-rgb, 0.35);
 
-@mixin outlinedColor($color) {
-	color: rgba($color, 1);
+@mixin setButtonColor($color: '', $color-hovered: '', $color-focused: '', $color-disabled: '') {
+	@if $color != '' {
```
kvas-damian, line 11: you could get rid of `ifs` if you use `null` instead of empty string: https://blog.teamtreehouse.com/smarter-sass-mixins-null

lib/styles/settings/_buttons.scss
damianmef: I replaced all `transparent` to correct token f.e.: `$color-primary-background-ghost-disabled'`

lib/styles/components/_buttons.scss
damianmef: It is, I'm replacing it to `var(--raw-black-rgb)`

lib/styles/components/_buttons.scss
damianmef: Done.

lib/styles/components/_buttons.scss
damianmef: Every usage of this mixin have all parameters, so I'm removing optional values.

lib/styles/components/_buttons.scss
damianmef: Yes, the effect is the same and we get rid of important.

lib/styles/components/_buttons.scss
damianmef: Ok, I removed `a-iconButton` from this PR and will make changes in `IconButton` in the next PR.

lib/styles/components/_buttons.scss
damianmef: Generally I'm removing default styles because we cant use component without color from the list. So situation when color will be not defined will not occur.

lib/styles/components/_buttons.scss
damianmef: It's good news. I replaced all '' to null in `Button`.

lib/styles/settings/_colors.scss
```diff
@@ -1,89 +1,82 @@
...
-$color-secondary-rgb: var(--color-secondary-rgb);
-$color-secondary-supporting-rgb: var(--color-secondary-supporting-rgb);
-$color-total-white-opacity50: rgba(255, 255, 255, 0.5);
+@import '../../styles/settings/colors/raw';
+
+$color-firefly-black: var(--raw-black);
+$color-firefly-black-rgb: var(--raw-black-rgb);
+$color-total-white: var(--raw-white);
+
+$color-wasabi-green: var(--raw-green-200);
```
kvas-damian, line 7: what about these colors? Didn't we agreed to revert them until stage 4?

lib/styles/components/_buttons.scss
```diff
@@ -3,48 +3,177 @@
 @import '../settings/spacings';
 @import '../settings/radiuses';
 @import '../settings/typography';
+@import '../settings/buttons';
 
-$default-lighter-shadow: 0 4px 12px rgba($color-firefly-black, 0.35);
```
kvas-damian, line 7: do we need to import old `colors` in this file?

lib/styles/settings/_buttons.scss
```diff
@@ -1,4 +1,5 @@
 @import 'colors';
```
kvas-damian, line 1: old colors are still imported

lib/styles/components/_buttons.scss
kvas-damian: you changed default value to `null` but `ifs` are still in place. Do you want to keep them in the code for readibility?

lib/styles/settings/_buttons.scss
damianmef: Fixed. @kvas-damian temporary i need to comment styles in `IconButton` when I removed old structures/colors from button in `IconButton` were errors. Whole `IconButton` will be replaced in the next PR

lib/styles/settings/_colors.scss
damianmef: Fixed, I edded removed unused colours.

lib/styles/components/_buttons.scss
damianmef: Having `if` in the code we generating less final code. I know is very small improvement. I have no problem with these `if`.

lib/styles/components/_buttons.scss
damianmef: Fixed.

lib/styles/components/_buttons.scss
damianmef: @kvas-damian I removed redundant `if` conditions. By the way, I changed the `BadgeScore`. I know it doesn't fit into this PR, but I didn't want to create a separate one for such a small thing.

lib/styles/settings/_colors.scss
kvas-damian: I'm not sure what's fixed. I still see these colors on a diff in PR, so they are changed. Why?

damianmef: You mean I should revert:
```
$color-wasabi-green: var(--raw-green-200);
```
to:
```
$color-wasabi-green: #aed581;
```
? I don't see any point in taking it back

kvas-damian: We talked about it at least 2 times already. We agreed that we move this change into stage 4. I don't undestand why you push on this now.

This change isn't backwards compatible. You can't pass CSS variables into `mix(), darken(), lighten()` SASS functions. Example: https://github.com/bethinkpl/wnl-platform/blob/9e34e6e69bec18340297af2c35b665f969108a89/resources/assets/js/components/global/Sidenav.vue#L33

damianmef: Right. File reverted @kvas-damian

kvas-damian (approved):

PR #117, opened 2022-04-10
Colors migration - 2c - Migrate IconButton

damianmef:


lib/js/components/Buttons/IconButton/IconButton.vue
```diff
@@ -50,42 +57,83 @@
...
+			#{$self}__label {
+				color: map-get($color-map, 'outlined', 'color');
+			}
 
-	@each $color-name, $color in $regular-colors {
-		&.-color-#{$color-name} {
-			@include iconButtonColor($color);
+			&:hover,
+			&.-hovered {
+				#{$self}__button {
```
damianmef, line 71: To check by me, is `background-color` and `color` necessary. Maybe it could be replaced by styling `__icon` instead of `__button`

lib/js/components/Buttons/IconButton/IconButton.vue
```diff
@@ -18,7 +23,14 @@
 		<div
 			v-if="$slots.default && type !== ICON_BUTTON_TYPES.ICON_ONLY"
 			class="a-iconButton__label"
-			:class="{ '-minor': colorScheme === ICON_BUTTON_COLOR_SCHEMES.MINOR_LABEL }"
+			:class="{
+				'-neutral':
+					colorScheme === ICON_BUTTON_COLOR_SCHEMES.NEUTRAL_LABEL &&
+					color !== ICON_BUTTON_COLORS.NEUTRAL_WEAK,
```
kvas-damian, line 29: why do we exclude `netural-weak` here?

lib/js/components/Buttons/IconButton/IconButton.vue
```diff
@@ -50,42 +57,83 @@
...
+				}
+				#{$self}__label {
+					color: map-get($color-map, 'filled', 'color-focused');
+				}
+			}
+
+			&.-disabled {
+				#{$self}__button {
+					background-color: map-get($color-map, 'filled', 'background-disabled');
+					color: map-get($color-map, 'icon');
```
kvas-damian, line 104: do we need it? It's the same as default color

lib/js/components/Buttons/IconButton/IconButton.vue
```diff
@@ -50,42 +57,83 @@
...
+
+					&.-outlined {
+						background-color: map-get($color-map, 'outlined', 'background');
```
kvas-damian, line 107: do we need to override original value here?

lib/js/components/Buttons/IconButton/IconButton.vue
```diff
@@ -50,42 +57,83 @@
...
 	align-items: center;
-	color: $color-primary;
+	color: $color-primary-text;
 	cursor: pointer;
 	display: inline-flex;
 	transition: color ease-in-out $default-transition-time;
 
+	&:disabled,
+	&.-disabled {
+		cursor: not-allowed;
```
kvas-damian, line 131: it won't work, because we disabled `pointer-events`

lib/js/components/Buttons/IconButton/IconButton.vue
```diff
@@ -113,14 +160,53 @@
 			display: initial;
 		}
 
-		&.-minor {
-			@include iconButtonColor($color-minor);
+		&.-neutral {
+			color: $color-neutral-text !important;
```
kvas-damian, line 164: @damianmef I looked at it. You could exclude these classes in the loop, but I'd also choose `important`

lib/js/components/Buttons/IconButton/IconButton.consts.ts
```diff
@@ -29,3 +28,10 @@ export const ICON_BUTTON_COLORS = {
 	WARNING: 'warning',
 	INFO: 'info',
 } as const;
```
damianmef, line 30: > after this change story for IconButton won't work correctly. I see multiple usages of BUTTON_COLORS.MINOR there
by Damian J.

lib/js/components/Buttons/IconButton/IconButton.vue
damianmef: 1. I think it's not worth to make changes. The problem is that class `hovered` is above `a-button` in the HTML tree. So if we want to move `__button` resposibility to `Button` it will more complicated.
2. Changing to the `__icon` will not fix it, because icon is inside. We need change background of the button (wrapper of icon). In above code we change `a-iconButton__button` element, not `a-button` component. It's the same HTML tag with classes:
```
class="a-button a-iconButton__button -outlined -color-primary"
```

lib/js/components/Buttons/IconButton/IconButton.vue
damianmef: True, we don't need it.

lib/js/components/Buttons/IconButton/IconButton.vue
damianmef: Yes, because background in button has color (same color as here icon and border), we are set here ghost background (white).

lib/js/components/Buttons/IconButton/IconButton.vue
damianmef: It's not my idea, but right this rule do nothing and it's redundant. I will remove it.

lib/js/components/Buttons/IconButton/IconButton.consts.ts
damianmef: In this PR there is no `minor` word in `components/Buttons` so it's outdated comment I think.

lib/styles/components/_icon-button.scss
```diff
@@ -0,0 +1,232 @@
...
+		.a-iconButton {
+			&__icon,
+			&__loadingIcon {
+				color: $icon;
+			}
+		}
+	}
+}
+
+.a-iconButton {
```
kvas-damian, line 35: we extracted styles for `buttons` into a separate file only becasue Button is used outside of Vue components - in Blade files. I don't see any reason to follow this rule here. It's easier to read and maintain code when styles and HTML are in the same file

lib/js/components/Buttons/IconButton/IconButton.vue
damianmef: I'm also not a fan of `important` and use it in specially situations, where it saves a lot of code and probably will not interfere with other rules in the future. If any `important` will make problems in the future, it should be rebuild. Ok?

lib/styles/components/_icon-button.scss
damianmef: Ok let it be, it's not so important now for me to keep it in separated file.

lib/js/components/Buttons/IconButton/IconButton.vue
kvas-damian: does UX plan to use `NEUTRAL_LABEL` with `NEUTRAL_WEAK`? In figma I see `ALL_IN_COLOR` in such cases. If we agree that UX uses `ALL_IN_COLOR`, then we get no exception in the code => better maintenability

lib/js/components/Buttons/IconButton/IconButton.vue
```diff
@@ -49,42 +56,120 @@
...
+			}
 		}
 	}
 
 	align-items: center;
-	color: $color-primary;
+	color: $color-primary-text;
 	cursor: pointer;
 	display: inline-flex;
 	transition: color ease-in-out $default-transition-time;
```
rogatty, line 164: Something seems to be overriding this transition time. It feels much longer on master.

lib/js/components/Buttons/IconButton/IconButton.vue
damianmef: According to Figma: https://www.figma.com/file/izQdYyiBR1GQgFkaOIfIJI/LMS---DS---Components?node-id=2770%3A62766
All cases with gray label has always three tokens: `$color-neutral-text`, `$color-neutral-text-hovered` , `$color-neutral-text-focused`, `$color-neutral-text-disabled`, `neutral-weak` is outdated. So condition looks simple now:
```
:class="{
	'-neutral': colorScheme === ICON_BUTTON_COLOR_SCHEMES.NEUTRAL_LABEL,
}"
```

kvas-damian:
probably you should merge `master` into previous branches. Otherwise we get a lot of unrelated changes on diff

lib/js/components/Buttons/IconButton/IconButton.vue
```diff
@@ -49,117 +51,231 @@
...
+
+		@media #{breakpoint-s()} {
+			display: initial;
+		}
+
+		&.-neutral {
+			color: $color-neutral-text !important;
+		}
+
+		&.-neutral-weak {
```
kvas-damian, line 198: is it still used?

damianmef: Nope. Updated.

kvas-damian (approved):

PR #120, opened 2022-04-11
IT-5005 | Vue 3

rogatty:
https://bethink.atlassian.net/browse/IT-4121
https://bethink.atlassian.net/browse/IT-5005
https://github.com/bethinkpl/wnl-platform/pull/3739


lib/js/components/Banner/Banner.vue
```diff
@@ -28,50 +25,38 @@
...
-							class="ds-banner__rightSlot"
-						>
+						<div v-if="$slots.rightSlot" class="ds-banner__rightSlot">
 							<slot name="rightSlot" />
 						</div>
 					</div>
 				</div>
 			</div>
-			<div
-				v-if="$slots.expandedText && $slots.expandedText.length > 0"
```
kvas-damian, line 45: vue3 return slots as function, so we can't check `length`. Fixed `v-if` on template in stories

.storybook/main.js
```diff
@@ -9,7 +9,31 @@ module.exports = {
...
+	core: {
+		builder: 'webpack5',
+	},
 	webpackFinal: async (config) => {
+		config.resolve.alias.vue = '@vue/compat';
+
+		let vueLoaderRule = config.module.rules.find(
+			(r) =>
+				// it can be another rule with file loader
+				// we should get only svg related
```
rogatty, line 21: This comment doesn't fit.

lib/js/components/Buttons/Button/Button.vue
```diff
@@ -21,7 +21,6 @@
 			'-elevation-x-small': elevation === ELEVATIONS.X_SMALL,
 			'-elevation-small': elevation === ELEVATIONS.SMALL,
 		}"
-		@click="$emit('click')"
```
rogatty, line 24: Will we have to update all button uses in platform (`@click.native`)? Is it necessary because of https://github.com/vuejs/core/issues/813? We could always use `emits: ["click"]` but maybe it's better to go straight for the best solution.

lib/js/components/SelectList/SelectList.stories.ts
```diff
@@ -2,16 +2,18 @@ import SelectList from './SelectList.vue';
...
 
-const StoryTemplate: StoryFn<typeof SelectList> = (argTypes) => ({
+const StoryTemplate: StoryFn<typeof SelectList> = (args) => ({
 	components: { SelectList },
-	props: Object.keys(argTypes),
+	setup() {
+		return { ...args };
+	},
 	template: '<select-list v-model="value" :items="items" :size="size"></select-list>',
 	data() {
```
rogatty, line 18: Something is wrong here:

Screenshot 2023-01-27 at 16 08 49

lib/js/components/Buttons/Button/Button.vue
kvas-damian: after upgrade, `.native` modifier is ignored and all events are handled by `@click`. By removing this line, we don't emit duplicated event in this component, so migration in platform should be limited to `@click.native => @click`

lib/js/components/SelectList/SelectList.stories.ts
rogatty: There is a lot of warnings about it in test logs, too.

rogatty: `icon` prop passed to https://github.com/bethinkpl/design-system/blob/0f5267f6737eb1a363656c4cedaed9a2d386c75a/lib/js/components/Icons/Icon/Icon.vue#L36-L38 is of a `Proxy` type, so that's probably the reason it doesn't validate properly.
Screenshot 2023-02-10 at 17 07 07

rogatty:
Nothing more to be done, we're waiting for platform migration before this can be merged.

kvas-damian (approved):

PR #122, opened 2022-04-13
Fixes reorganization structure after merge

damianmef:


kvas-damian:
we agreed that we don't commit `/dist` to avoid conflicts

lib/js/components/Cards/Card/index.ts
```diff
@@ -0,0 +1,3 @@
+import Card from './Card.vue';
```
kvas-damian, line 1: I don't understand why we add these index files if we don't use them. Do you plan to use them in `lib/js/index.ts` or should they be removed?

damianmef: After discussion: All imports with `.vue` in `vue` files are replaced by import folder which contains `index.ts` file.

PR #126, opened 2022-04-20
IT-4442 | Add Divider component

rogatty:
https://bethink.atlassian.net/browse/IT-4442
https://www.figma.com/file/izQdYyiBR1GQgFkaOIfIJI/LMS---DS---Components?node-id=1552%3A34960
https://miro.com/app/board/uXjVO_rTPmE=/?moveToWidget=3458764522565102566&cot=14
https://wnl-platform-dev-kuba.s3.eu-central-1.amazonaws.com/public/storybook/IT-4442/index.html?path=/story/components-divider--interactive


lib/js/components/Divider/Divider.stories.mdx
```diff
@@ -0,0 +1,36 @@
...
+		   prominence: {
+			   control: { type: 'select', options: Object.values(PROMINENCES) },
+			   defaultValue: PROMINENCES.DEFAULT,
+		   },
+	   }}>
+	{(args) => {
+		return {
+			components: { Divider },
+			props: Object.keys(args),
+			template: '<div style="height: 200px; width: 200px;"><divider :is-vertical="isVertical" :prominence="prominence"/></div>',
```
rogatty, line 29: @damianmef @kvas-damian do we have any agreements on how to present components in a pretty way? Current solution works but looks bad.

Also, do we need a separate story for each variant? I see we do that for some of the components but don't know if it's a rule.

damianmef: @rogatty Yes, we have similar situation for contrast background (it's not yet in master):
https://github.com/bethinkpl/design-system/blob/23702055d956516c67a3f2c25a3577cecb0750fa/lib/styles/storybook.scss
just add class in story and css in `storybook.scss`

I don't know findings about separated stories, but it looks clear if we have many stories. You can click from menu every variant and component code is more clear. Additionally you see values of properties and you can send url for specific variant.

lib/js/components/Divider/Divider.stories.mdx
```diff
@@ -0,0 +1,36 @@
+import { Meta, Story, Props } from '@storybook/addon-docs';
+import Divider from './Divider'; import { PROMINENCES } from './Divider.consts';
```
damianmef, line 2: Do You have prettier turned on? It should move import to the next line.

damianmef:
@rogatty please updatee `js/index.ts` adding Divider. Before that add `index.ts` inside `Divider` folder. We are not more using includes with `.vue` suffix.

lib/js/components/Divider/Divider.stories.mdx
kvas-damian: > Also, do we need a separate story for each variant? I see we do that for some of the components but don't know if it's a rule.
Variants are usually requested by UX, I don't see this requirement for Divider: https://miro.com/app/board/uXjVO_rTPmE=/?moveToWidget=3458764522565713965&cot=10 but you can confirm that

lib/js/components/Divider/Divider.vue
```diff
@@ -0,0 +1,59 @@
...
+
+<script lang="ts">
+import { PROMINENCES } from './Divider.consts';
+
+export default {
+	name: 'Divider',
+	props: {
+		isVertical: {
+			type: Boolean,
+			required: true,
```
kvas-damian, line 47: According to https://miro.com/app/board/uXjVO_rTPmE=/?moveToWidget=3458764522565713964&cot=10, this should be optional

lib/js/components/Divider/Divider.vue
```diff
@@ -0,0 +1,59 @@
+<template>
+	<div
```
kvas-damian, line 2: Did you try using `<hr>`? It's more semantically correct

lib/js/components/Divider/Divider.vue
```diff
@@ -0,0 +1,59 @@
...
+		height: 1px;
+		width: 100%;
+	}
+}
+</style>
+
+<script lang="ts">
+import { PROMINENCES } from './Divider.consts';
+
+export default {
```
kvas-damian, line 42: you forgot to export it in https://github.com/bethinkpl/design-system/blob/master/lib/js/index.ts

lib/js/components/Divider/Divider.vue
rogatty: Even if it's vertical? 🙂

kvas-damian: you can change the `tagName` based on props ;)

lib/js/components/Divider/Divider.stories.mdx
rogatty: I did not have it turned on. Fixed, thanks.

lib/js/components/Divider/Divider.vue
rogatty: To what? There is no `<vr>` 😛  `<hr>` was my first instinct too, but it doesn't fit vertical line. Also:
>The \<hr> HTML element represents a thematic break between paragraph-level elements: for example, a change of scene in a story, or a shift of topic within a section.

I'm not sure if it's even a good match for how we'll be using it.

lib/js/index.ts
```diff
@@ -7,6 +7,7 @@ export { default as Button } from './components/Buttons/Button';
 export * from './components/Buttons/Button/Button.consts';
 export { default as Card } from './components/Cards/Card/';
 export { default as CardExpandable } from './components/Cards/CardExpandable/';
+export * from './components/Divider';
```
kvas-damian, line 10: does it export `default`?

rogatty: https://github.com/bethinkpl/design-system/blob/a6c4c40bd80acb563c97186584974c4eee298d44/lib/js/components/Divider/index.ts#L1-L4

kvas-damian (approved):

PR #127, opened 2022-04-22
Add cache to github actions

damianmef:


.github/workflows/build.yml
```diff
@@ -9,6 +9,10 @@ jobs:
     steps:
       - name: Checkout 🛎️
         uses: actions/checkout@v2
+      - uses: actions/cache@v2
+        with:
+          path: '**/node_modules'
+          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
```
kvas-damian, line 15: Why do we use two different techniques for caching? Can we unify them?

damianmef: I was following the platform. One cache is typical for node modules, second is for `setup-node@v2` to be honest I did not analyze the difference, I assumed that this is deliberate in the platform.

.github/workflows/ci.yml
```diff
@@ -19,6 +19,7 @@ jobs:
       - uses: actions/setup-node@v2-beta
         with:
           node-version: '16.13'
+          cache: 'yarn'
```
kvas-damian, line 22: https://github.com/bethinkpl/design-system/actions/runs/2206635938/attempts/1

```
Warning: Unexpected input(s) 'cache', valid inputs are ['always-auth', 'node-version', 'architecture', 'check-latest', 'registry-url', 'scope', 'token', 'version']
```

.github/workflows/build.yml
kvas-damian: IMO, it's a mistake in `wnl-platform` that we shouldn't follow

damianmef: Ok.

kvas-damian: Now, we don't use any cache in `ci.yaml`. What I was forcing is: cache unification across different workflows in this repo, not removal of the cache

kvas-damian: Also, I suggest using https://github.com/actions/setup-node#caching-global-packages-data - it's dedicated for Node, so it's possible github optimized it for this use-case

kvas-damian:
> if CI is working with actions @v3 instead of @v2 can we bump it?

I don't see any reason to freeze it at the old version. So good to go!

.github/workflows/ci.yml
```diff
@@ -13,14 +13,15 @@ jobs:
...
+      - uses: actions/checkout@v3
       - name: Insert auth font awesome auth token
         run: echo "@fortawesome:registry=https://npm.fontawesome.com/" > .npmrc && echo "//npm.fontawesome.com/:_authToken=${{ secrets.FONT_AWESOME_AUTH_TOKEN }}" >> .npmrc
-      - uses: actions/setup-node@v2-beta
+      - uses: actions/setup-node@v3
         with:
           node-version: '16.13'
+          cache: 'yarn'
       - name: Install dependencies
-        run: yarn install --pure-lockfile
```
kvas-damian, line 23: why did we remove this parameter?

PR #131, opened 2022-04-27
Colors migration 2.0 fixes

damianmef:


lib/js/components/Buttons/IconButton/IconButton.vue
```diff
@@ -31,7 +31,13 @@
 		<wnl-button
 			ref="button"
 			class="a-iconButton__button"
-			:class="{ '-iconOnly': type === ICON_BUTTON_TYPES.ICON_ONLY }"
+			:class="{
+				'-iconOnly': type === ICON_BUTTON_TYPES.ICON_ONLY,
+				'-hovered': state === ICON_BUTTON_STATES.HOVERED,
+				'-focused': state === ICON_BUTTON_STATES.FOCUSED,
+				'-loading': state === ICON_BUTTON_STATES.LOADING,
```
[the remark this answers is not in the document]
kvas-damian: IMO, this key isn't defined in `ICON_BUTTON_STATES`. Also, exposing state which isn't implmented could be confusing for future readers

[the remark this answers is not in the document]
damianmef: True, loading removed.

lib/js/components/Buttons/IconButton/IconButton.vue
```diff
@@ -95,48 +99,26 @@
...
 			&.-disabled {
 				#{$self}__button {
-					background-color: map-get($color-map, 'filled', 'background-disabled');
-
 					&.-outlined {
-						background-color: map-get($color-map, 'outlined', 'background');
-						border-color: map-get($color-map, 'outlined', 'disabled', 'border');
-
 						.a-iconButton__icon {
 							color: map-get($color-map, 'outlined', 'disabled', 'icon');
```
damianmef, line 118: Finally we are not touching `__button` but need touch this class, because icon in IconButton in Button is different element than icon in Button component.

lib/styles/components/_buttons.scss
```diff
@@ -291,21 +291,31 @@ $button-disabled-alpha: 0.6;
...
+				transform: translateZ(-1px);
 			}
 		}
-	}
 
-	&.-elevation-small {
-		&.-outlined {
-			&:not(.-hovered):not(:hover):not(.-focused):not(:focus):not(.-disabled) {
-				background: $color-default-background !important;
+		&.-rounded {
```
kvas-damian, line 311: we use `::before` only for `-outlined`. For safety, I'd move this block inside `&.-outlined {` block

kvas-damian (approved):

lib/styles/components/_buttons.scss
damianmef: @kvas-damian i fixed in in the wrong branch: https://github.com/bethinkpl/design-system/blob/b5fb1bfd3cc51cb25f902a949aa68a48751b44de/lib/styles/components/_buttons.scss#L220

PR #135, opened 2022-05-02
IT-4481 shadows migration

damianmef:


lib/styles/settings/_shadows.scss
```diff
@@ -1,4 +1,13 @@
...
+$color-shadow: rgba(var(--raw-black-rgb), 0.3);
+$color-success-shadow: rgba(var(--raw-green-500-rgb), 0.7);
+$color-danger-shadow: rgba(var(--raw-red-500-rgb), 0.7);
+$color-warning-shadow: rgba(var(--raw-gold-500-rgb), 0.7);
+
+$color-neutrals-shadow-weak: rgba(var(--raw-black-rgb, 0.06));
+$color-neutrals-shadow: rgba(var(--raw-black-rgb, 0.08));
+$color-neutrals-shadow-strong: rgba(var(--raw-black-rgb, 0.12));
 
 $shadow-s: 0 1px 2px $color-neutrals-shadow-weak, 0 8px 16px $color-neutrals-shadow-weak,
```
kvas-damian, line 12: these shadows still use old colors

lib/styles/settings/_colors.scss
```diff
@@ -101,16 +101,6 @@ $color-fail: $color-outrageous-orange;
 $color-fail-supporting: $color-melon-orange;
 $color-fail-background: $color-ice-orange;
 
-$color-shadow: rgba($color-firefly-black, 0.35);
```
kvas-damian, line 104: did you think about consequences of this change? Is it backwards compatible? Do you need to adapt platform code?

lib/js/components/SurveyQuestions/SurveyQuestionTextarea.vue
```diff
@@ -20,7 +20,7 @@
 
 	border: 1px solid $color-mischka-gray;
 	box-sizing: border-box;
-	box-shadow: inset 0 1px 3px $color-minor-supporting;
+	box-shadow: inset 0 1px 3px var(--raw-gray-400);
```
kvas-damian, line 23: do we need to import old `colors` after this change?

lib/styles/settings/_shadows.scss
damianmef: `$color-neutrals-shadow-weak` is defined a few lines above, it uses new colors.

lib/styles/settings/_colors.scss
damianmef: @kvas-damian I moved it to `settings/_shadows.scss` I just decided to move all shadow settings to one file.

lib/js/components/SurveyQuestions/SurveyQuestionTextarea.vue
damianmef: Yes, because there is still colors from this file in usage. f.e. `$color-mischka-gray` Of course usages will disappear in the `colors-migration-2.0`.

kvas-damian: so after we merge both branches, we will end up with not used import. Do you have a plan how to fix that?

lib/styles/settings/_colors.scss
kvas-damian: I understand that, but I'm worried if it won't break platform when you install this version.

PR #155, opened 2022-06-24
IT-4616 | Pill

kvas-damian:
Additionally:
* new `IconButton` size: `XX_SMALL`
* new `IconButton` color: `FAIL`
* new `Icon` size: `XXX_SMALL`

### DIFF
--- a/lib/js/components/Pill/Pill.vue
+++ b/lib/js/components/Pill/Pill.vue
@@ -86,7 +86,6 @@ $pill-colors: (
 	}
 
 	align-items: center;
-	background-color: $color-primary-background-strong;
 	border-radius: 100px;
 	display: inline-flex;
 	padding: $space-xxxxxs $space-xxxxxs $space-xxxxxs $space-xxs;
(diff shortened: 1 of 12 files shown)


lib/js/components/Pill/Pill.vue
```diff
@@ -0,0 +1,196 @@
...
+		:title="label"
+	>
+		<icon
+			v-if="leftIcon"
+			class="pill__leftIcon"
+			:icon="leftIcon"
+			:size="size === PILL_SIZES.X_SMALL ? ICON_SIZES.XXX_SMALL : ICON_SIZES.XX_SMALL"
+		/>
+		<span class="pill__label">{{ label }}</span>
+		<icon-button
```
pioziol123, line 14: Is it alway the same size? Also there is no deletable small pill version in figma. Should we allow for that?

kvas-damian: Good catch, asked UX: https://www.figma.com/file/HFY6waq7kWsr3rj8a0ghXZ?node-id=1499:118929#217941952

kvas-damian: UX confirmed that we should hide it on `x-small`: [`831f5b0` (#155)](https://github.com/bethinkpl/design-system/pull/155/commits/831f5b0c257749759f541fab5e944ffebc6016d5)

PR #168, opened 2022-07-21
IT-4661 | Expose Banner expanded state

rogatty:
Needed for https://bethink.atlassian.net/browse/IT-4661 / https://github.com/bethinkpl/wnl-platform/pull/3872

Mostly copied from https://github.com/bethinkpl/design-system/blob/88e51d6aa0466e39d07dbb8b57a3633b0d65868d/lib/js/components/Cards/CardExpandable/CardExpandable.vue

### DIFF
--- a/lib/js/components/Banner/Banner.stories.mdx
+++ b/lib/js/components/Banner/Banner.stories.mdx
@@ -22,8 +22,10 @@ import { ICONS } from '../Icon';
 		title: 'Banner Title',
 		buttonText: null,
 		layout: BANNER_LAYOUTS.HORIZONTAL,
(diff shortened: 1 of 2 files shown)


lib/js/components/Banner/Banner.stories.mdx
```diff
@@ -43,7 +45,7 @@ import { ICONS } from '../Icon';
 		},
 		buttonText: {
 			control: { type: 'text' },
-			defaultValue: null,
+			defaultValue: '',
```
kvas-damian, line 48: this prop is still rendered with red border. I think you have to update line 23.

kvas-damian (approved):

lib/js/components/Banner/Banner.stories.mdx
rogatty: Works for me: 
image

rogatty: Actually, you were right, fixed in https://github.com/bethinkpl/design-system/commit/51bf6610124fc787057cd845103342d4b79381bf

PR #170, opened 2022-07-22
Fix: remove old theme-variables

kvas-damian:
old theme css variables are conflicting with css variables for new tokens: https://github.com/bethinkpl/design-system/blob/2dbaef0922a7fbd459cbdb0f672fe646a7c35836/lib/styles/theme-varaibles.scss#L12
Let's clean it up


lib/styles/components/_links.scss
```diff
@@ -1,9 +1,11 @@
+@import '../settings/colors/tokens';
+
 .a-link {
 	cursor: pointer;
-	color: $color-primary;
+	color: $color-primary-text;
```
rogatty, line 5: @kvas-damian it was `#0f9698` and now it's `#088289`. I don't think it's correct.

kvas-damian: there is no other primary color token for text

rogatty: UX is ok with that.

rogatty (approved):

PR #171, opened 2022-08-01
IT-4682 | Add SelectList component

rogatty:
- https://bethink.atlassian.net/browse/IT-4682
- https://www.figma.com/file/izQdYyiBR1GQgFkaOIfIJI/LMS---DS---Components?node-id=3375%3A57750
- https://miro.com/app/board/uXjVOZ_Oix8=/?moveToWidget=3458764529299372340&cot=14


lib/js/components/SelectList/SelectList.vue
```diff
@@ -0,0 +1,133 @@
...
+	<div class="selectList">
+		<div
+			v-for="(item, index) in items"
+			:key="index"
+			class="selectList__item"
+			:class="getClassNamesForItem(item)"
+			@click="onItemClick(item)"
+		>
+			<divider v-if="item.type === 'divider'" />
+			<template v-else-if="item.type === 'text'">
```
tweetgeek, line 11: if someone use an invalid type, we display an empty div, is that good behavior?

rogatty: It's on purpose. Do you have a better alternative? You don't know what fields will be available.

lib/js/components/SelectList/SelectList.vue
```diff
@@ -0,0 +1,133 @@
+<template>
+	<div class="selectList">
+		<div
+			v-for="(item, index) in items"
+			:key="index"
```
tweetgeek, line 5: i know that we have many indexes as key, but i'm not a fan of this solution, look to https://vueschool.io/articles/vuejs-tutorials/tips-and-gotchas-for-using-key-with-v-for-in-vue-js-3/ at "No indexes as keys"

lib/js/components/SelectList/SelectList.vue
```diff
@@ -0,0 +1,133 @@
...
+			},
+		},
+		value: {
+			type: String,
+			default: null,
+		},
+	},
+	created() {
+		this.ICON_SIZES = ICON_SIZES;
+		this.ICONS = ICONS;
```
tweetgeek, line 106: we also don't use ICONS :)

lib/js/components/SelectList/SelectList.vue
tweetgeek: maybe we should check in the validator of this prop if someone passed `text` or `divider` items? It seems to be the simplest way

rogatty: Validator works only in dev mode and we already have TS helping devs. I'm not sure if yet another validation step gives much value. We'll have to keep this code anyway for production.

tweetgeek: theoretically, you are right, but dev should know that something has been misplaced - when it goes into production, it's too late :)
or maybe we should do something like this
```
<template>
	<div class='selectList'>
		<template v-for='(item, index) in items'>
			<div
				v-if="AVAILABLE_TYPES.includes(item.type)"
				:key='index'
				class='selectList__item'
				:class='getClassNamesForItem(item)'
				@click='onItemClick(item)'
			>
				<divider v-if="item.type === 'divider'" />
				<template v-else-if="item.type === 'text'">
					<ds-icon
						v-if='item.icon'
						class='selectList__itemIcon'
						:icon='item.icon'
						:size='ICON_SIZES.X_SMALL'
					/>
					{{ item.label }}
				</template>
			</div>
		</template>
	</div>
</template>
```
then we don't render an empty div

tweetgeek: on the other hand, maybe that's not a problem, this div has `height: 0px`

rogatty: Dev will know the data format is wrong if they use the exported interface, no? It's always hard to draw the line on how deffensive we want to be. I'd say the current state is fine because we give all the tools needed and don't crash the component if you make a mistake. Worst case scenario you get an empty div with zero height 🙂

PR #179, opened 2022-08-17
Clean up - remove old colors - Stage 4

damianmef:


lib/styles/modifiers/_colors.scss
```diff
@@ -1,35 +1,31 @@
...
 
-.-fireflyBlack {
-	color: $color-firefly-black;
-}
-
-.-delugeViolet {
-	color: $color-deluge-violet;
+.-violet500 {
+	color: var(--raw-violet-500);
 }
```
kvas-damian, line 31: I believe 1 class is missing. Was it intended?

damianmef: This class has no usage in DS and in Platform.

damianmef: @kvas-damian what do You think if we will remove all these classes? I will be better `clean up`. I checked usages in the Platform:
```
-gray400 - 5
-gray700 - 19
-gray300 - 4
-white - 0?
-green500 - 4
-blue500 - 0
-black  - 0
-violet500 - 0
```
All cases: 32.

I think we should replace like this:
```
<td class="-gray400">
```
to:
```
<td class="grayCell"> 
and
.gray { var(--raw-gray-400); }
```

lib/styles/settings/_shadows.scss
```diff
@@ -1,6 +1,15 @@
-@import 'colors';
 @import 'colors/tokens';
 
+$color-shadow: rgba(var(--raw-black-rgb), 0.3);
+$color-success-shadow: rgba(var(--raw-green-500-rgb), 0.7);
+$color-danger-shadow: rgba(var(--raw-red-500-rgb), 0.7);
+$color-warning-shadow: rgba(var(--raw-gold-500-rgb), 0.7);
+
+// Tokens - shadows
+$color-neutrals-shadow-weak: $color-default-shadow-weak;
```
kvas-damian, line 9: why do we have to create an alias on this variable? Isn't original name good enough?

damianmef: I think I didn't want to touch shadows, it was before shadows migration. Now these variables have no usages, so I removed them.

lib/styles/modifiers/_colors.scss
damianmef: File removed.

kvas-damian (approved):

PR #180, opened 2022-08-24
IT-4617 | PopOver and Dropdown

kvas-damian:
https://miro.com/app/board/uXjVOZ_Oix8=/?moveToWidget=3458764526448763872&cot=14

Styles based on:
* https://github.com/bethinkpl/wnl-platform/blob/02c0ced665cc3e40fbe8757869574eb1fbf93c76/resources/assets/js/components/global/PopperDropdown.vue
* https://github.com/bethinkpl/wnl-platform/blob/1b5982c53778a140028e05135f157f523fd41696/resources/assets/js/components/global/Popper.vue


lib/js/components/Buttons/Button/Button.vue
```diff
@@ -21,6 +21,7 @@
 			'-elevation-x-small': elevation === ELEVATIONS.X_SMALL,
 			'-elevation-small': elevation === ELEVATIONS.SMALL,
 		}"
+		@click="$emit('click')"
```
kvas-damian, line 24: this makes it easier to bind to `click` event. We don't need to add `.native` modifier anymore

lib/js/components/Badges/BadgeScore/BadgeScore.vue
```diff
@@ -120,14 +120,14 @@ export default {
 			type: String,
 			required: true,
 			validate(color) {
-				Object.values(BADGE_SCORE_COLORS).includes(color);
+				return Object.values(BADGE_SCORE_COLORS).includes(color);
```
kvas-damian, line 123: `validate()` method has to return value, otherwise it's useless

lib/js/components/PopOver/PopOver.vue
```diff
@@ -0,0 +1,222 @@
...
+	data() {
+		return {
+			POP_OVER_COLORS: Object.freeze(POP_OVER_COLORS),
+			BUTTON_TYPES: Object.freeze(BUTTON_TYPES),
+			BUTTON_SIZES: Object.freeze(BUTTON_SIZES),
+			key: 1,
+		};
+	},
+	watch: {
+		triggerAction() {
```
pioziol123, line 201: Maybe we can make it a mixin?

pioziol123 (approved):

lib/js/components/PopOver/PopOver.vue
kvas-damian: we have different prop names to be watched in these components. Having a mixin just for a single property watch is an overkill. 
Do you have alternative solution in your mind?

pioziol123: No. Just popped into my mind :)

PR #189, opened 2022-09-20
MDX to TS init

damianmef:


lib/js/components/Badges/Badge/Badge.stories.ts
```diff
@@ -0,0 +1,39 @@
...
+} as Meta<typeof Badge>;
+
+const Template: Story<{ label: string; iconUrl: string; color: Object }> = (argTypes) => ({
+	components: { Badge },
+	props: Object.keys(argTypes),
+	template:
+		'<div style="display: flex"><Badge :color="color" :label="label"><img alt="Badge" style="width: 100%; height: 100%" :src="iconUrl" /></Badge></div>',
+});
+
+export const Primary = Template.bind({});
```
kvas-damian, line 20: why do we call it `Primary`?

lib/js/components/Badges/Badge/Badge.stories.ts
```diff
@@ -0,0 +1,39 @@
...
+import { BADGE_COLORS } from './Badge.consts';
+
+import { Meta, Story } from '@storybook/vue';
+
+export default {
+	title: 'Components/Badges/Badge',
+	component: Badge,
+} as Meta<typeof Badge>;
+
+const Template: Story<{ label: string; iconUrl: string; color: Object }> = (argTypes) => ({
```
kvas-damian, line 13: is it ok that we call it `Template` while it's `Story` type?

lib/js/components/Icon/Icon.consts.ts
```diff
@@ -18,7 +18,7 @@ const BETHINK_ICONS = {
 	RIBBON: Ribbon,
 } as const;
 
-export const ICONS = {
+export let ICONS = {
```
kvas-damian, line 21: why this change?

lib/js/components/Badges/Badge/Badge.stories.ts
```diff
@@ -0,0 +1,39 @@
+// Badge.stories.ts
+
```
kvas-damian, line 2:
```suggestion
```

lib/js/components/Badges/Badge/Badge.stories.ts
```diff
@@ -0,0 +1,39 @@
...
+		defaultValue: BADGE_COLORS.NEUTRAL,
+	},
+};
+
+Primary.parameters = {
+	design: {
+		type: 'figma',
+		url: 'https://www.figma.com/file/izQdYyiBR1GQgFkaOIfIJI/LMS---DS---Components?node-id=1552%3A34953',
+	},
+};
```
kvas-damian, line 39: Can we remove `Badge.stories.mdx` now?

lib/js/components/Buttons/Button/Button.stories.ts
```diff
@@ -0,0 +1,104 @@
+// Button.stories.ts
+
```
kvas-damian, line 2:
```suggestion
```

lib/js/components/Buttons/Button/Button.stories.ts
```diff
@@ -0,0 +1,104 @@
...
+import { Meta, Story } from '@storybook/vue';
+
+export default {
+	title: 'Components/Buttons/Button',
+	component: Button,
+} as Meta<typeof Button>;
+
+const Template: Story<{
+	slotText: Object;
+	size: Object;
```
kvas-damian, line 23: incorrect type for many props in this list

lib/js/components/Buttons/Button/Button.stories.ts
```diff
@@ -0,0 +1,104 @@
...
+	components: { Button },
+	props: Object.keys(argTypes),
+	template: `<div :class="{ contrastBackground: isInverted }">
+                <Button
+                  :size="size"
+                  :type="type"
+                  :state="state"
+                  :radius="radius"
+                  :color="color"
+                  :elevation="elevation"
```
kvas-damian, line 41: did you consider using `v-bind=$props` syntax? I'm not sure what are dowsnsides of it, but it looks promising

lib/js/components/Buttons/Button/Button.stories.ts
```diff
@@ -0,0 +1,104 @@
...
+import { ICONS } from '../../Icon';
+
+import { Meta, Story } from '@storybook/vue';
+
+export default {
+	title: 'Components/Buttons/Button',
+	component: Button,
+} as Meta<typeof Button>;
+
+const Template: Story<{
```
kvas-damian, line 21: I see in docs that Storybook recommends `StoryFn()`: https://storybook.js.org/docs/vue/writing-stories/introduction
What was the reason to not follow that rule?

lib/js/components/Badges/Badge/Badge.stories.ts
```diff
@@ -0,0 +1,39 @@
...
+
+Primary.argTypes = {
+	color: {
+		control: { type: 'select', options: Object.values(BADGE_COLORS) },
+		defaultValue: BADGE_COLORS.NEUTRAL,
+	},
+};
+
+Primary.parameters = {
+	design: {
```
kvas-damian, line 35: IMO, there should be only one Design link per comonent, so it should be moved to `Story` definition: `const Template: Story<{ label: string; iconUrl: string; color: Object }> = (argTypes) => ({`

lib/js/components/Badges/Badge/Badge.stories.ts
kvas-damian, line 13: why `color` is `Object` type?

lib/js/components/Badges/Badge/Badge.stories.ts
damianmef: It should be `Interactive`.

lib/js/components/Badges/Badge/Badge.stories.ts
damianmef: Hmm so how we should name it? Like this?
```
const InteractiveStory: Story<....
...
export const Interactive = InteractiveStory.bind({});

```

kvas-damian: maybe `StoryTemplate` or `AbstractStory` from OOP?

lib/js/components/Icon/Icon.consts.ts
damianmef: I was sure it was suggestion by IDE, but now I haven't it. Will revert it.

lib/js/components/Badges/Badge/Badge.stories.ts
damianmef: It was suggestion in SB tutorials to add it, but it do nothing. So we can remove it.

lib/js/components/Buttons/Button/Button.stories.ts
damianmef: I think this is unacceptable. I will try find out other way.

lib/js/components/Badges/Badge/Badge.stories.ts
damianmef: `StoryTemplate` sounds well.

lib/js/components/Buttons/Button/Button.stories.ts
damianmef: Btw. This icon in `Badge` I think we can put for now like this:
```
src="https://lek.wiecejnizlek.pl/images/lek/logo-badge.svg" 
```
this will be replaced anyway in Badge refactoring tomorrow.

damianmef: @kvas-damian I found a way, but we need discuss is it profitable to keep such a code:
```
template: `<div :class="{ contrastBackground: isInverted }">
                <Button
			v-bind="propsWithoutSlot"
                  	:icon-left="ICONS[iconLeft]"
                  	:icon-right="ICONS[iconRight]"
                >
                    {{slotText}}
                </Button>
            </div>`,
	computed: {
		propsWithoutSlot() {
			const all = Object.assign({}, this.$props);
			delete all.slotText
			return all;
		}
	},
```

lib/js/components/Badges/Badge/Badge.stories.ts
kvas-damian: it's there only to tell you in which file you're, because in web code snippets you usually don't see filename

lib/js/components/Buttons/Button/Button.stories.ts
```diff
@@ -0,0 +1,97 @@
...
+} as Meta<typeof Button>;
+
+const StoryTemplate: StoryFn<{
+	slotText: Object;
+	size: String;
+	type: String;
+	color: String;
+	radius: String;
+	state: String;
+	elevation: String;
```
kvas-damian, line 26: https://stackoverflow.com/a/14727461/17570299

lib/js/components/Buttons/Button/Button.stories.ts
kvas-damian: it's not worth it. We can stay with the old version of passing props

lib/js/components/Badges/Badge/Badge.stories.ts
```diff
@@ -0,0 +1,41 @@
...
+
+const StoryTemplate: StoryFn<{ label: string; iconUrl: string; color: string }> = (argTypes) => ({
+	components: { Badge },
+	props: Object.keys(argTypes),
+	template:
+		'<div style="display: flex"><Badge v-bind=$props><img alt="Badge" style="width: 100%; height: 100%" :src="iconUrl" /></Badge></div>',
+	design: {
+		type: 'figma',
+		url: 'https://www.figma.com/file/izQdYyiBR1GQgFkaOIfIJI/LMS---DS---Components?node-id=1552%3A34953',
+	},
```
kvas-damian, line 19:
```suggestion
```

lib/js/components/Buttons/Button/Button.stories.ts
```diff
@@ -0,0 +1,97 @@
...
+
+import { Meta, StoryFn } from '@storybook/vue';
+
+export default {
+	title: 'Components/Buttons/Button',
+	component: Button,
+} as Meta<typeof Button>;
+
+const StoryTemplate: StoryFn<{
+	slotText: Object;
```
kvas-damian, line 20: I'm wondering if it's ok that we pass obejct with props as a type. In docs I see different approach: https://storybook.js.org/docs/vue/writing-stories/introduction#using-args

damianmef: Hmm it wasn't simple, I need to define `Args` and `ArgTypes` types of variables, @kvas-damian check code. But it works fine.

lib/js/components/Badges/Badge/Badge.stories.ts
damianmef: Removed.

lib/js/components/Buttons/Button/Button.stories.ts
damianmef: @kvas-damian my God, it looks nice but throwed me back to the problem:
Screenshot 2022-09-20 at 15 33 28

kvas-damian: is it related to the change of return type in `StoryFn`?

damianmef:
@kvas-damian I checked a few popular Storybooks and my thoughts are:
- they are not using slots, i have never seen `slot`
- half of them was not interactive
- buttons were always filled by `label` - input text

I checked platform usages of `<wnl-button>` and for first glance = we are using slot only for text. Maybe it's a good move to change it to props? I know it's time-consuming and we don't need to do that now, but I think it's a right way. It will also help to not prescribe props now and use it as now `v-bind=$props`.
Maybe we should turning a blind eye to these ugly properties in DS?

damianmef:
@kvas-damian I think I found a solution, in `Button.stories.ts`:
Slot as:
```
<template v-if="${'default' in args}" v-slot>${args.default}</template>
```
It has advantage - it appears as slot (screen). Disadvantage is that in html of component there is property `default="default"` but it doesn't contain slot content.
Screenshot 2022-09-27 at 18 24 26

kvas-damian:
> Maybe it's a good move to change it to props
UX already initiated trend to avoid slots in components, bu it isn't always doable. We need them on container components, like Modal, PopOver, etc, so we need a way to tackle them

> I think I found a solution, in Button.stories.ts
I'm not sure how it solves the problem, you still get slot arg defined on html element where you use `v-bind=$props`

kvas-damian:
I see that GitLab doesn't use `v-bind` syntax: https://gitlab.com/gitlab-org/gitlab-ui/-/blob/main/src/components/base/modal/modal.stories.js - maybe we should revert that for now?

damianmef:
@kvas-damian sure, I revereted pros. But I think it's a good idea to stay with default slot. I think we can now close PR and think about `$props` more in the future.

lib/js/components/Buttons/Button/Button.stories.ts
```diff
@@ -0,0 +1,97 @@
...
+	props: Object.keys(argTypes),
+	template: `<div :class="{ contrastBackground: isInverted }">
+                <Button
+					:size="size"
+					:type="type"
+					:state="state"
+					:radius="radius"
+					:color="color"
+					:elevation="elevation"
+                  	:icon-left="ICONS[$props.iconLeft]"
```
kvas-damian, line 30: tabs vs spaces

kvas-damian (approved):

PR #190, opened 2022-09-20
Fix icon button console log

damianmef:


damianmef:
@kvas-damian I tried to not duplicate colors like in this PR, but all attempts were unsuccessful. The main problem is we are using `Button` in `IconButton` and if I included `IconButton` colors in `Button` it was duplicated = error. `Button` need to know about additional `IconButton` colors like `info` that was the reason of errors in console. It didn't know about `info` color, having only it's own colors.

lib/js/components/Buttons/Button/Button.consts.ts
```diff
@@ -18,6 +18,14 @@ export const BUTTON_COLORS = {
 	INVERTED: 'inverted',
 } as const;
 
+export const BUTTON_COLORS_WITH_ICON_COLORS = {
```
kvas-damian, line 21: creating separate const can lead to misconsistencies. What'd you say about using `ICON_BUTTON_COLORS` to validate props?

lib/js/components/Buttons/Button/Button.vue
```diff
@@ -147,6 +148,7 @@ export default {
 		this.SIZES = BUTTON_SIZES;
 		this.STATES = BUTTON_STATES;
 		this.ELEVATIONS = BUTTON_ELEVATIONS;
+		this.BUTTON_COLORS_WITH_ICON_COLORS = BUTTON_COLORS_WITH_ICON_COLORS;
```
kvas-damian, line 151: https://bethinkteam.slack.com/archives/C01N1F1QMUY/p1650525683804989?thread_ts=1650477770.866669&cid=C01N1F1QMUY

In `platform` code we already started migration to `data()`

damianmef: Oh nice, so I refactored it by the way in `Button` and `IconButton` in this PR.

lib/js/components/Buttons/Button/Button.consts.ts
damianmef: Thanks to the fixing import path:
```
import { BUTTON_TYPES } from '../Button/Button.consts';
```
this additional const is redundant. I always clicked suggestion by IDE to make short path in import and that makes errors.

kvas-damian (approved):

PR #192, opened 2022-09-22
BadgeScore refactoring

damianmef:


lib/js/components/Badges/BadgeScore/BadgeScore.vue
```diff
@@ -20,98 +27,141 @@
...
+		#{$self}__suffix {
+			@include headlineXS();
+		}
 	}
 
 	&.-xsmall {
-		@include textXS();
-
+		font-size: $font-size-info-m;
+		line-height: $line-height-info-m;
```
damianmef, line 137: @kvas-damian we have no mixin for so small font in DS, is it worth to add the new one?

lib/js/components/Badges/BadgeScore/BadgeScore.stories.ts
```diff
@@ -0,0 +1,82 @@
...
+};
+
+/* FULL WIDTH */
+const StaticStoryTemplate: StoryFn<{
+	isFullWidth: boolean;
+}> = (argTypes) => ({
+	components: { BadgeScore },
+	props: Object.keys(argTypes),
+	template:
+		'<div v-if="isFullWidth" class="flexWrapper">' +
```
damianmef, line 67: @kvas-damian generally we should not use conditions in stories, but in this component we are adding additional wrapper, so this one condition is better than double whole `StoryFn` for only one class.

lib/js/components/Badges/BadgeScore/BadgeScore.vue
damianmef: My bad, it exists.

lib/js/components/Badges/BadgeScore/BadgeScore.stories.ts
kvas-damian: we discussed it off-line, probably there is no need for a wrapper after we change the way we control `width` of this component

lib/js/components/Badges/BadgeScore/BadgeScore.vue
```diff
@@ -20,98 +27,141 @@
...
+	components: {
+		WnlIcon,
+	},
 	props: {
 		text: {
 			type: String,
 			required: true,
 		},
-		additionalText: {
+		suffix: {
```
kvas-damian, line 164: This change is backwards incompatible. Do we have a PR in platform to address that?

lib/js/components/Badges/BadgeScore/BadgeScore.vue
```diff
@@ -20,98 +27,141 @@
...
-	border-width: 2px;
-	border-style: solid;
 	display: inline-block;
+	flex: 1;
 	min-width: $badge-score-width;
-	padding: $space-xxxs $space-xxxxs;
-	text-align: center;
+	padding: $space-xxxs $space-xxs;
+
+	&__wrapper {
```
kvas-damian, line 108: do we need additional wrapper? Maybe we could move these styles to `.badgeScore`?

lib/js/components/Badges/BadgeScore/BadgeScore.vue
```diff
@@ -7,11 +7,18 @@
...
 			'-neutral': color === BADGE_SCORE_COLORS.NEUTRAL,
+			'-neutral-medium': color === BADGE_SCORE_COLORS.NEUTRAL_MEDIUM,
 			'-small': size === BADGE_SCORE_SIZES.SMALL,
 			'-xsmall': size === BADGE_SCORE_SIZES.XSMALL,
 		}"
 	>
-		{{ text }}<span class="badgeScore__additionalText">{{ additionalText }}</span>
+		<div class="badgeScore__wrapper">
+			<wnl-icon v-if="icon" class="badgeScore__icon" :icon="icon" :size="iconSize" />
+			<div class="badgeScore__content">
```
kvas-damian, line 17: instead of separate `div` you could apply `align-self: baseline` on text and suffix

lib/js/components/Badges/BadgeScore/BadgeScore.vue
```diff
@@ -123,6 +173,14 @@ export default {
 				return Object.values(BADGE_SCORE_COLORS).includes(color);
 			},
 		},
+		icon: {
+			type: Object,
+			required: false,
```
kvas-damian, line 181: not needed when you pass `default` value

lib/js/components/Badges/BadgeScore/BadgeScore.vue
```diff
@@ -20,98 +27,141 @@
 @import '../../../../styles/settings/typography';
 @import '../../../../styles/settings/colors/tokens';
 
-$badge-score-width: 74px;
-$small-badge-score-width: 48px;
-$x-small-badge-score-width: 36px;
+$badge-score-width: 66px;
```
kvas-damian, line 30: it's `min-width`, we should change variable name

lib/js/components/Badges/BadgeScore/BadgeScore.stories.ts
damianmef: I think it's the simplest way, minus of this solution is component has unused class:
```
	&.-fullWidth {
		width: 100%;
	}
```
but it's only one rule and we know that this component can be dependent of parent node.

@kvas-damian I still dont know the magic of `min-width` with flex, maybe You mean this example solution:
```
.container .box {
  min-width: 200px;
  max-width: 400px;
  flex-basis: auto; /* default value */
  flex-grow: 1;
}
```
Anyway we don't need it.

lib/js/components/Badges/BadgeScore/BadgeScore.stories.ts
```diff
@@ -0,0 +1,79 @@
...
+	components: { BadgeScore },
+	props: Object.keys(argTypes),
+	template:
+		'<badge-score :color="BADGE_SCORE_COLORS.WARNING" suffix="%" text="100" :class="{ \'-fullWidth\': isFullWidth }" />',
+	created() {
+		this.BADGE_SCORE_COLORS = BADGE_SCORE_COLORS;
+	},
+});
+
+export const MinWidth = StaticStoryTemplate.bind({});
```
kvas-damian, line 73: do we want to add a note for dev that they should use `min-width` to set width of this component when needed?
Another option is to pass `min-width` value through props.

lib/js/components/Badges/BadgeScore/BadgeScore.vue
damianmef: We haven't yet, but I think it's good time to change this property. We decided to name it `suffix` it's more precisely. I decided to do that also of small cases of usage - 5 in Platform. I can prepare PR in platform after when we finished this PR.

lib/js/components/Badges/BadgeScore/BadgeScore.vue
damianmef: We can, but then it always will be `flex` not `display-inline`. I'm not sure we want it flex as default. What do You think, knowing places where it occurs?

lib/js/components/Badges/BadgeScore/BadgeScore.vue
damianmef: We have there `align-items: baseline;` on parent, so true, it's redundant.

lib/js/components/Badges/BadgeScore/BadgeScore.stories.ts
damianmef: Is not better way to use `-fullWidth` class and set the width of parent in used place? It would not interfere with the component

lib/js/components/Badges/BadgeScore/BadgeScore.vue
kvas-damian: you can use `inline-flex`

lib/js/components/Badges/BadgeScore/BadgeScore.vue
damianmef: @kvas-damian I created subtask for this.

lib/js/components/Badges/BadgeScore/BadgeScore.stories.ts
kvas-damian: what if UX want it to be `min-width: 200px`?

lib/js/components/Badges/BadgeScore/BadgeScore.vue
damianmef: Ok, it works fine.

lib/js/components/Badges/BadgeScore/BadgeScore.vue
kvas-damian: I still see this container in HTML

lib/js/components/Badges/BadgeScore/BadgeScore.vue
kvas-damian: remember, that we need it ready before we release DS

lib/js/components/Badges/BadgeScore/BadgeScore.vue
```diff
@@ -20,98 +24,141 @@
...
-		padding: $space-xxxxs;
+		min-width: $badge-score-x-small-min-width;
+		padding: $space-xxxs;
+
+		#{$self}__suffix {
+			@include textInfoM();
+		}
+	}
+
+	&.-full-width {
```
kvas-damian, line 140: we try to use camel case for class names

lib/js/components/Badges/BadgeScore/BadgeScore.vue
```diff
@@ -20,98 +24,141 @@
...
 
 	border-radius: 4px;
-	border-width: 2px;
-	border-style: solid;
-	display: inline-block;
-	min-width: $badge-score-width;
-	padding: $space-xxxs $space-xxxxs;
-	text-align: center;
+	display: inline-flex;
+	flex: 1;
```
kvas-damian, line 101: is it still needed?

damianmef: Removed.

lib/js/components/Badges/BadgeScore/BadgeScore.vue
damianmef: Updated.

kvas-damian (approved):

PR #193, opened 2022-09-26
IT-4784 | add rightSlot to Banner

tweetgeek:


lib/js/components/Banner/Banner.vue
```diff
@@ -104,6 +120,14 @@
...
 	&.-vertical {
+		#{$self}__rightWrapperHorizontal {
+			display: none;
+		}
+
+		#{$self}__rightWrapperVertical {
+			display: flex;
+		}
+
 		#{$self}__buttonTextHorizontal {
```
kvas-damian, line 131: do we still need it?

lib/js/components/Banner/Banner.vue
```diff
@@ -249,18 +306,15 @@
 
 	&__buttonTextHorizontal {
 		padding: $space-xs 0;
-		flex-shrink: 0;
 		display: none;
```
kvas-damian, line 309: do we need to manipulate with childs `display` while we hide/show wrapper?

lib/js/components/Banner/Banner.vue
```diff
@@ -197,6 +221,39 @@
 		padding: $space-xxxxs 0;
 	}
 
+	&__rightWrapperHorizontal {
```
kvas-damian, line 224: don't we need to vertically align items in this wrapper. When I type some text, it's positioned higher than button

kvas-damian: the same for vertical

tweetgeek: I changed the vertical/horizontal display, no duplicate code any more [`89e1e65` (#193)](https://github.com/bethinkpl/design-system/pull/193/commits/89e1e657b51a65c9a777bab702e67133da5ecaae)

lib/js/components/Banner/Banner.vue
```diff
@@ -104,76 +93,44 @@
 	padding: $space-xs;
 
 	&.-vertical {
```
kvas-damian, line 95: not used now

kvas-damian, line 95: the same styles have to applied on mobile breakpoint

lib/js/components/Banner/Banner.vue
```diff
@@ -189,24 +146,45 @@
...
 		display: flex;
 		flex-direction: column;
 		justify-content: center;
 		flex-grow: 1;
 		padding: $space-xxxxs 0;
 	}
 
-	&__iconContainer {
-		@include iconContainerVerticalStyles;
+	&__rightWrapper {
```
kvas-damian, line 164: items in this container aren't aligned

lib/js/components/Banner/Banner.vue
```diff
@@ -7,33 +7,34 @@
...
-					<div v-if="buttonText" class="ds-banner__buttonTextVertical">
-						<ds-button
-							:color="BUTTON_COLORS.NEUTRAL"
-							:type="BUTTON_TYPES.OUTLINED"
-							:size="BUTTON_SIZES.SMALL"
-							@click.native="$emit('button-clicked')"
-							>{{ buttonText }}
-						</ds-button>
+					<div class="ds-banner__rightWrapper">
+						<div v-if="buttonText" class="ds-banner__buttonText">
```
kvas-damian, line 21: do we need a wrapper on button? It looks that you could simplify it and move class name on Button

damianmef: If we need it should be named as other wrappers `ds-banner_buttonWrapper`

lib/js/components/Banner/Banner.vue
```diff
@@ -291,6 +244,41 @@
 	&__expandedText {
 		margin-top: $space-xs;
 	}
+
+	&.-warning {
```
damianmef, line 248: I think modifying classes of main component class should be after it's rules, not at the end after children rules.

lib/js/components/Banner/Banner.vue
kvas-damian: @tweetgeek I see the wrapper is still here. DO you want to share argument behind this decision?

lib/js/components/Banner/Banner.vue
kvas-damian: @tweetgeek still valid

tweetgeek: its like in figma, wrapper has correct paddings, alignment to top of this wrapper

lib/js/components/Banner/Banner.vue
tweetgeek: its gone

lib/js/components/Banner/Banner.vue
kvas-damian: Because in Figma, there is only one example with component that has the same height as the Button. You can put some test or smaller component to see the problem in DS. It's possible that bug is also present in Figma

tweetgeek: I think that I don't understand what the problem is? Ewa said it was to be aligned to the top with specific paddings, just like in figma, and that's what it is made. Should we make it dependent on the size of the component putted in the slot?

kvas-damian: I'm not sure where Ewa said that, maybe she didn't analyze all possible cases?
See the screenshot when I put 15x15px square into this slot. For me it looks off
Zaznaczenie_916

lib/js/components/Banner/Banner.vue
```diff
@@ -189,24 +183,45 @@
...
 		flex-direction: column;
 		justify-content: center;
 		flex-grow: 1;
 		padding: $space-xxxxs 0;
 	}
 
-	&__iconContainer {
-		@include iconContainerVerticalStyles;
+	&__rightWrapper {
+		display: flex;
```
kvas-damian, line 202: becuase of `flex`, button is growing when text goes into multiple lines
Zaznaczenie_917

lib/js/components/Banner/Banner.vue
tweetgeek: Ok i understand, I asked about it https://bethinkteam.slack.com/archives/C012R8RM3A4/p1664458749846619?thread_ts=1663751823.320519&cid=C012R8RM3A4

lib/js/components/Banner/Banner.vue
tweetgeek: fiexed, so wrapper are required :D

kvas-damian (approved):

lib/js/components/Banner/Banner.vue
rogatty: Will be addressed in another PR.

PR #194, opened 2022-09-28
Progress bar

damianmef:


lib/js/components/ProgressBar/ProgressBar.vue
```diff
@@ -0,0 +1,330 @@
...
+			'-info': color === PROGRESS_BAR_COLORS.INFO,
+			'-schemeMedium': colorScheme === PROGRESS_BAR_COLOR_SCHEMES.MEDIUM,
+			'-schemeMediumNeutral': colorScheme === PROGRESS_BAR_COLOR_SCHEMES.MEDIUM_NEUTRAL,
+			'-compact': layout === PROGRESS_BAR_LAYOUTS.COMPACT,
+		}"
+	>
+		<div class="progressBar__label">
+			<div class="progressBar__labelText">{{ labelText }}</div>
+			<div v-if="labelDataExist" class="progressBar__labelDataWrapper">
+				<div v-if="labelData" class="progressBar__labelData">{{ labelData }}</div>
```
kvas-damian, line 19: because label is rendered as inline text, `span` looks more semantic for me

lib/js/components/ProgressBar/ProgressBar.vue
```diff
@@ -0,0 +1,330 @@
...
+		},
+		size: {
+			type: String,
+			default: PROGRESS_BAR_SIZES.MEDIUM,
+			validate(size) {
+				return Object.values(PROGRESS_BAR_SIZES).includes(size);
+			},
+		},
+		ranges: {
+			type: Array,
```
kvas-damian, line 281: Do we want to defined PropType for range to make it easier to guess what are required attributes here?

lib/js/components/ProgressBar/ProgressBar.vue
```diff
@@ -0,0 +1,330 @@
...
+$progress-bar-height: 16px;
+$progress-bar-s-height: 8px;
+$progress-bar-xs-height: 4px;
+
+$progress-bar-border-radius: 8px;
+
+$progress-bar-layers: (
+	'primary': (
+		'default': $color-primary-text,
+		'ghost': $color-primary-data-ghost,
```
kvas-damian, line 68: instead of naming it based on color used, I'd name it based on usage, eg. `default-secondary-layer`. If you do that, there won't be any problems when someone defines non-ghoust color here. What do you think?

lib/js/components/ProgressBar/ProgressBar.vue
```diff
@@ -0,0 +1,330 @@
...
+		}
+		#{$self}__label {
+			margin-bottom: $space-xxxs;
+		}
+	}
+
+	&__bar {
+		background-color: $color-default-background;
+		border-radius: $progress-bar-border-radius;
+		box-shadow: inset 0 1px 4px $color-default-shadow-heavy;
```
kvas-damian, line 161: If I remember correctly UX wanted this shadows to be defined as tokens, so probably they should be accessible in other components. Am I wrong?

lib/js/components/ProgressBar/ProgressBar.vue
```diff
@@ -0,0 +1,330 @@
...
+		height: $progress-bar-height;
+		overflow: hidden;
+
+		&.-small {
+			height: $progress-bar-s-height;
+		}
+
+		&.-xsmall {
+			border-radius: 0;
```
kvas-damian, line 170: where did we agree on disabling radius for this size?

lib/js/components/ProgressBar/ProgressBar.vue
```diff
@@ -0,0 +1,330 @@
...
+		</div>
+		<div
+			:class="{
+				progressBar__bar: true,
+				'-small': size === PROGRESS_BAR_SIZES.SMALL,
+				'-xsmall': size === PROGRESS_BAR_SIZES.XSMALL,
+				'-noRadius': radius === PROGRESS_BAR_RADII.NONE,
+			}"
+		>
+			<div class="progressBar__results">
```
kvas-damian, line 37: do we need this container? Why don't we define styles on parent?

lib/js/components/ProgressBar/ProgressBar.vue
```diff
@@ -0,0 +1,330 @@
...
+
+					&.-secondary {
+						background: map-get($color-map, 'neutral-weak');
+					}
+				}
+			}
+		}
+	}
+
+	&-noRadius {
```
kvas-damian, line 145: I see the similar block with `.` at the beginning few lines below. Should we remove it?

lib/js/components/ProgressBar/ProgressBar.vue
```diff
@@ -0,0 +1,330 @@
...
+	PROGRESS_BAR_LAYOUTS,
+	PROGRESS_BAR_COLOR_SCHEMES,
+} from './ProgressBar.consts';
+
+export default {
+	name: 'ProgressBar',
+	props: {
+		numberOfLayers: {
+			type: Number,
+			default: 1,
```
kvas-damian, line 257: Do you want to use const (already defined) as default value? do we want to validate this value?

lib/js/components/ProgressBar/ProgressBar.vue
```diff
@@ -0,0 +1,330 @@
...
+		colorScheme: {
+			type: String,
+			default: PROGRESS_BAR_COLOR_SCHEMES.DEFAULT,
+			validate(size) {
+				return Object.values(PROGRESS_BAR_COLOR_SCHEMES).includes(size);
+			},
+		},
+		color: {
+			type: String,
+			required: true,
```
kvas-damian, line 268: According to Miro, `info` is default color

lib/js/components/ProgressBar/ProgressBar.consts.ts
```diff
@@ -0,0 +1,35 @@
...
+	FAIL: 'fail',
+} as const;
+
+export const PROGRESS_BAR_SIZES = {
+	MEDIUM: 'medium',
+	SMALL: 'small',
+	XSMALL: 'extra small',
+} as const;
+
+export const PROGRESS_BAR_RADII = {
```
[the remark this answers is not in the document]
damianmef: Plural of `radius`, `radiuses` is underlined by the IDE, `radii` not.

lib/js/components/ProgressBar/ProgressBar.vue
damianmef: Ok, I will do that.

lib/js/components/ProgressBar/ProgressBar.vue
```diff
@@ -0,0 +1,330 @@
...
+		>
+			<div class="progressBar__results">
+				<div
+					v-for="(range, index) in ranges"
+					:key="index"
+					:class="{
+						progressBar__result: true,
+						'-secondary': range.layer === 2 && numberOfLayers === 2,
+					}"
+					:style="{ left: range.percentValueFrom + '%', width: range.length + '%' }"
```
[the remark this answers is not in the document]
damianmef: @kvas-damian hmm I think `start` does not fit. Maybe just:
```
export interface Ranges {
	layer: Number,
	from: Number,
	length: Number,
}
```

lib/js/components/ProgressBar/ProgressBar.vue
damianmef: It's good idea, but it's hard to hame these keys, what do you think about these keys:
```
		'default-layer': $color-primary-text,
		'default-layer-secondary-': $color-primary-data-ghost,
		'second-layer': $color-primary-data-medium,
		'second-layer-secondary': $color-primary-data-weak,
		'placeholder-layer': $color-neutral-data-weak,
```

lib/js/components/ProgressBar/ProgressBar.vue
damianmef: Im not sure, in Figma this element has `default/shadowProgress -M` so yes it's a token, I can move it to `_shadows` where we have other tokens, but this token has so specified name so it will be used only in this component.

lib/js/components/ProgressBar/ProgressBar.vue
damianmef: True, it's a mistake.

lib/js/components/ProgressBar/ProgressBar.vue
damianmef: I will remove it.

lib/js/components/ProgressBar/ProgressBar.vue
damianmef: Fixed

lib/js/components/ProgressBar/ProgressBar.vue
damianmef: Fixed.

lib/js/components/ProgressBar/ProgressBar.vue
```diff
@@ -0,0 +1,330 @@
...
+
+$progress-bar-height: 16px;
+$progress-bar-s-height: 8px;
+$progress-bar-xs-height: 4px;
+
+$progress-bar-border-radius: 8px;
+
+$progress-bar-layers: (
+	'primary': (
+		'default': $color-primary-text,
```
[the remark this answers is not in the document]
damianmef: My bad, fixed.

lib/js/components/ProgressBar/ProgressBar.vue
```diff
@@ -0,0 +1,330 @@
...
+		return {
+			PROGRESS_BAR_COLORS: Object.freeze(PROGRESS_BAR_COLORS),
+			PROGRESS_BAR_SIZES: Object.freeze(PROGRESS_BAR_SIZES),
+			PROGRESS_BAR_RADII: Object.freeze(PROGRESS_BAR_RADII),
+			PROGRESS_BAR_LAYOUTS: Object.freeze(PROGRESS_BAR_LAYOUTS),
+			PROGRESS_BAR_COLOR_SCHEMES: Object.freeze(PROGRESS_BAR_COLOR_SCHEMES),
+		};
+	},
+	computed: {
+		labelDataExist() {
```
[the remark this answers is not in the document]
damianmef: Fixed.

lib/js/components/ProgressBar/ProgressBar.vue
damianmef: Fixed.

lib/js/components/ProgressBar/ProgressBar.consts.ts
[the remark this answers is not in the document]
kvas-damian: https://dictionary.cambridge.org/pl/dictionary/english/radii

lib/js/components/ProgressBar/ProgressBar.vue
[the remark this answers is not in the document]
damianmef: It convinced me.

lib/js/components/ProgressBar/ProgressBar.vue
kvas-damian: IMO, proposed key names are wrong:
`second-layer` uses color that is used for 1st data layer in `MEDIUM` color scheme

Please try again

lib/js/components/ProgressBar/ProgressBar.vue
kvas-damian: what about other `divs` inside `__labelDataWrapper`?

lib/js/components/ProgressBar/ProgressBar.consts.ts
```diff
@@ -0,0 +1,41 @@
...
+	TWO: 2,
+};
+
+export const PROGRESS_BAR_COLOR_SCHEMES = {
+	DEFAULT: 'default',
+	MEDIUM: 'medium',
+	MEDIUM_NEUTRAL: 'medium-neutral',
+};
+
+export interface Ranges {
```
kvas-damian, line 37: this type defines one range. Why did you use plural in name?

lib/js/components/ProgressBar/ProgressBar.stories.ts
```diff
@@ -0,0 +1,174 @@
...
+		control: { type: 'select', options: Object.values(PROGRESS_BAR_LAYERS) },
+		defaultValue: PROGRESS_BAR_LAYERS.ONE,
+	},
+	colorScheme: {
+		control: { type: 'select', options: Object.values(PROGRESS_BAR_COLOR_SCHEMES) },
+		defaultValue: PROGRESS_BAR_COLOR_SCHEMES.DEFAULT,
+	},
+	color: {
+		control: { type: 'select', options: Object.values(PROGRESS_BAR_COLORS) },
+		defaultValue: PROGRESS_BAR_COLORS.PRIMARY,
```
kvas-damian, line 37: INFO is default color
```suggestion
		defaultValue: PROGRESS_BAR_COLORS.INFO,
```

lib/js/components/ProgressBar/ProgressBar.vue
damianmef: Also, I hope these are not the first symptoms of Divitis.

lib/js/components/ProgressBar/ProgressBar.consts.ts
damianmef: Fixed.

lib/js/components/ProgressBar/ProgressBar.stories.ts
damianmef: Fixed.

lib/js/components/ProgressBar/ProgressBar.vue
damianmef: When I look at this screen:
Screenshot 2022-09-30 at 11 45 37
and 3 layer of data, I dont see sense of changing these keys. If someone would use noghost color, he should add new key `noghost` and use it the new place or replace `ghost` by it.

lib/js/components/ProgressBar/ProgressBar.vue
damianmef: In progress.

lib/js/components/ProgressBar/ProgressBar.vue
kvas-damian: ok, let's try another game:
Imagine, we keep current colors config scheme. In 3 months UX wants us to change color for second layer in default color scheme from `ghost` to `strong`. What needs to be done?

damianmef: I understand it, but can't find correct word. Check below names, if they are wrong please help me:
```
'default-layer': $color-primary-data,
'default-layer-secondary': $color-primary-data-ghost,
'emphasize-layer': $color-primary-data-medium,
'emphasize-layer-secondary': $color-primary-data-weak,
'placeholder-layer': $color-neutral-data-weak,
```

lib/js/components/ProgressBar/ProgressBar.vue
damianmef: @kvas-damian I change it, shadows are now global and have global names, after consultation they looks like here: https://www.figma.com/file/CG4lkRM9SnX7Qjz6rZxLYZ/LMS---DS---Foundations?node-id=209%3A2868

lib/js/components/ProgressBar/ProgressBar.vue
kvas-damian: what would you say about
```
'default-color-scheme-first-layer`
'default-color-scheme-second-layer`
'medium-color-scheme-first-layer`
...
'medium-neutral-color-scheme-second-layer`
```
?

lib/js/components/ProgressBar/ProgressBar.vue
```diff
@@ -226,6 +231,15 @@ $progress-bar-layers: (
 		position: absolute;
 		top: 0;
 	}
+
+	&__resultsShadow {
```
kvas-damian, line 235: told ya, "use `background` for ranges, then eveything works". Includign shadows :trollface:

lib/js/components/ProgressBar/ProgressBar.vue
```diff
@@ -44,6 +44,7 @@
 				:style="{ left: range.percentValueFrom + '%', width: range.length + '%' }"
 			>
 			</div>
+			<div class="progressBar__resultsShadow"></div>
```
kvas-damian, line 47: you know, I don't non-meaningful, not required HTML markup. Can we get rid of it?

lib/js/components/ProgressBar/ProgressBar.vue
damianmef: Sounds good, but is `color-scheme` needed? It's only internal key for this variable in this single component. Without it is clear for me also. We haven't there any other colors so these are unambiguous for me.

kvas-damian: when I was yunior dev, I though I don't need to make variable names self explanatory - it's all visible in the code. Now I see, I was wrong. We read code much more often than we write it, so forcing someone to go through code once again to understand something is a waste of time. That's why I like longer names now.

lib/js/components/ProgressBar/ProgressBar.vue
damianmef: > you know, I don't non-meaningful, not required HTML markup. Can we get rid of it?
It's not clear for me. How You will add shadow to whole progressbar? On parent it doesn't work (layers are above), when I add to single layer, we have error https://bethinkteam.slack.com/archives/CFBLF5RGA/p1664803900641519?thread_ts=1664532308.131999&cid=CFBLF5RGA , this markup is a fix for this.

kvas-damian: did you try `::after`?

lib/js/components/ProgressBar/ProgressBar.vue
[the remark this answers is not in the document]
kvas-damian: so why it isn't implemented?

lib/js/components/ProgressBar/ProgressBar.vue
```diff
@@ -116,29 +116,29 @@ $progress-bar-layers: (
...
-						background: map-get($color-map, 'weak');
+						background: map-get($color-map, 'medium-color-scheme-second-layer');
 					}
 				}
 			}
 
 			&.-schemeMediumNeutral {
 				#{$self}__result {
-					background: map-get($color-map, 'medium');
+					background: map-get($color-map, 'medium-color-scheme-first-layer');
```
kvas-damian, line 138: I'm not fan of reusing color from other color scheme, but don't have strong arguments to convince you not to do that

lib/js/components/ProgressBar/ProgressBar.vue
damianmef: @kvas-damian of course it's possible, I was thinking about it and i chose HTML tag. I do not insist on him, but I also do not see any advantage in doing it by :after. What's your argument, shorter/clear HTML? Shorter HTML but more css rules.

lib/js/components/ProgressBar/ProgressBar.vue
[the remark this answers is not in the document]
damianmef: Now it is.

lib/js/components/ProgressBar/ProgressBar.vue
kvas-damian: every HTML element means additional additional DOM node in DOM tree. When the tree is big, traversing and updating it is more computing heavy. Pseudoelements aren't put in DOM, so they don't affect that area.

Additional argument is separation of code layers. When you use `::after` then styles belonging to an element are kept in one place, when you put them in additional HTML element, then the logic is distributed and have to be maintained in two places.

lib/js/components/ProgressBar/ProgressBar.stories.ts
```diff
@@ -0,0 +1,191 @@
...
+	numberOfLayers: 2,
+	ranges: [
+		{
+			layer: 1,
+			percentValueFrom: 0,
+			length: 10,
+		},
+		{
+			layer: 2,
+			percentValueFrom: 20,
```
kvas-damian, line 152: didn't we already change this field name?

To solve that problem you could use `Range` interface here

lib/js/components/ProgressBar/ProgressBar.vue
damianmef: Changed.

lib/js/components/ProgressBar/ProgressBar.stories.ts
damianmef: Updated.

PR #204, opened 2022-10-21
IT-4439 | Migrate deprecated slot syntax

rogatty:
It's required before we can try Vue 3 migration build: [v3-migration.vuejs.org/migration-build.html#preparations](https://v3-migration.vuejs.org/migration-build.html#preparations)


.eslintrc.js
```diff
@@ -1,97 +1,129 @@
...
-    }],
-    'vue/multi-word-component-names': 0
-  }
-};
\ No newline at end of file
+	plugins: ['import', '@typescript-eslint'],
+	extends: [
+		'eslint:recommended',
+		'plugin:vue/recommended',
+		'prettier',
```
rogatty, line 6: https://github.com/prettier/eslint-config-prettier/blob/main/CHANGELOG.md#version-800-2021-02-21

.eslintrc.js
```diff
@@ -1,97 +1,129 @@
...
+		'vue/no-potential-component-option-typo': 'error',
+		'vue/no-static-inline-styles': 'error',
+		'vue/no-template-target-blank': [
+			'error',
+			{
+				allowReferrer: true,
+			},
+		],
+		'vue/v-for-delimiter-style': 'error',
+		'vue/no-undef-components': 'error',
```
rogatty, line 126: https://eslint.vuejs.org/rules/no-unregistered-components.html
> 🚫 This rule was removed in eslint-plugin-vue v9.0.0 and replaced by [vue/no-undef-components](https://eslint.vuejs.org/rules/no-undef-components.html) rule.

lib/js/components/Buttons/Button/Button.vue
```diff
@@ -70,6 +70,7 @@ import {
 import { ICON_BUTTON_COLORS } from '../IconButton/IconButton.consts';
 
 export default {
+	// eslint-disable-next-line vue/no-reserved-component-names
 	name: 'Button',
```
rogatty, line 74: We'll fix it by adding `Ds` prefix which we have plans for.

lib/js/components/SurveyQuestions/SurveyQuestionScale/SurveyQuestionScale.spec.ts
```diff
@@ -29,10 +28,6 @@ describe('SurveyQuestionScale', () => {
 			slots: {
 				explanation,
 			},
-			stubs: {
-				SurveyToggle,
-				Card,
```
rogatty, line 34: Alternatively we could use string stubs, something like `<div><slot/><slot name="footer"/></div>`. Using `mount` was easier.

kvas-damian (approved):

PR #205, opened 2022-10-24
IT-3288 spinner loading

damianmef:


lib/js/components/Buttons/IconButton/IconButton.vue
```diff
@@ -372,6 +385,24 @@ export default {
 	},
 	computed: {
 		iconSize(): string {
+			if (this.state === ICON_BUTTON_STATES.LOADING) {
```
damianmef, line 388: It need have separated condition with return at the end.

lib/js/components/Buttons/IconButton/IconButton.vue
```diff
@@ -166,6 +173,12 @@
 		}
 	}
 
+	&.-loading {
+		.a-icon {
+			animation: a-fadeIn 300ms ease-in-out 0s 1, a-spinAround 1500ms infinite linear;
```
damianmef, line 178: I know we want to avoid styling child component, but here it's a content of slot and we are not touching Button's loading state. Button has 3 sizes, IconButton 5 and different sizes of icons. Thats why I didn't use Button state and added here second icon with condition state=loading.

damianmef: If it's acceptable, maybe I could move this definition of animation as some global variable.

damianmef:
@kvas-damian general info: in task there is info about add loading state to `CaptionButton` but it's not avaiable in the DS.

lib/js/components/Buttons/IconButton/IconButton.vue
```diff
@@ -43,7 +44,13 @@
...
 			:color="isButtonColor ? color : null"
 		>
-			<wnl-icon class="a-iconButton__icon" :icon="icon" :size="iconSize" />
+			<wnl-icon
+				v-if="state === ICON_BUTTON_STATES.LOADING"
+				class="a-iconButton__icon"
+				:icon="ICONS.FA_SPINNER_THIRD"
+				:size="iconSize"
+			/>
+			<wnl-icon v-else class="a-iconButton__icon" :icon="icon" :size="iconSize" />
```
kvas-damian, line 53: the only difference in this `v-if` is icon. Why don't render other icon conditionally in `:icon` prop?

lib/js/components/Buttons/IconButton/IconButton.vue
kvas-damian: Did you try using `spinning` prop on `Icon`? Maybe we'll need to adjust spin time, but IMO it's valuable to unify default spinning time.

If it isn't possible, why don't you rotate based on `.a-iconButton__icon` selector?

lib/js/components/Buttons/IconButton/IconButton.vue
kvas-damian: are you sure these sizes are different from `ICON_ONLY_ICON_SIZES_MAP`?

kvas-damian: BTW. Only `ICON_ONLY` type should have these sizes

lib/js/components/Buttons/Button/Button.vue
```diff
@@ -67,7 +67,11 @@ import {
 	BUTTON_TYPES,
 } from './Button.consts';
 
-import { ICON_BUTTON_COLORS } from '../IconButton/IconButton.consts';
+import {
+	ICON_BUTTON_COLORS,
+	ICON_BUTTON_SIZES,
+	ICON_BUTTON_TYPES,
```
kvas-damian, line 73: not used imports?

lib/js/components/Buttons/IconButton/IconButton.vue
damianmef: > BTW. Only ICON_ONLY type should have these sizes, see:

@kvas-damian are You sure? `Outlined` and `Filled` are using same icon and have same sizes (5 types). This loader should looks same for `iconOnly` and other variants.

damianmef: > are you sure these sizes are different from ICON_ONLY_ICON_SIZES_MAP?

Right! Condition should be:

```
			if (this.type === ICON_BUTTON_TYPES.ICON_ONLY || this.state === ICON_BUTTON_STATES.LOADING) {
				return ICON_ONLY_ICON_SIZES_MAP[this.size];
			}
```

lib/js/components/Buttons/IconButton/IconButton.vue
damianmef: Thanks, I missed this `spinning` property. It works well.

lib/js/components/Buttons/IconButton/IconButton.vue
damianmef: Now is the second difference - `spinning` - I think this condition should stay.

lib/js/components/Buttons/Button/Button.vue
damianmef: Yes, removed.

lib/js/components/Buttons/IconButton/IconButton.vue
kvas-damian: For `filled L` I see Icon XS: https://www.figma.com/file/izQdYyiBR1GQgFkaOIfIJI/LMS---DS---Components?node-id=4803%3A83076

lib/js/components/Buttons/IconButton/IconButton.vue
```diff
@@ -43,7 +44,14 @@
 			:elevation="elevation"
 			:color="isButtonColor ? color : null"
 		>
-			<wnl-icon class="a-iconButton__icon" :icon="icon" :size="iconSize" />
+			<wnl-icon
+				v-if="state === ICON_BUTTON_STATES.LOADING"
+				class="a-iconButton__icon"
+				:icon="ICONS.FA_SPINNER_THIRD"
+				:size="iconSize"
+				:spinning="true"
```
kvas-damian, line 52:
```suggestion
				spinning
```

kvas-damian, line 52:
```suggestion
				spinning
```

damianmef: Fixed.

damianmef: Fixed.

lib/js/components/Buttons/IconButton/IconButton.vue
damianmef: Fixed.

kvas-damian (approved):

PR #211, opened 2022-10-26
PoC - typography tokens

damianmef:


damianmef:
@kvas-damian I think we should add all rules to tokens. In the code was situation token was applied to `progressBar__labelDataWrapper` and it's child `progressBar__labelDataSuffix` also had token. Parent could have rule about `text-decoration: underline` and then child will inherit it. Maybe it's unreal scenario, but we need consider it.

lib/styles/settings/typography/_tokens.scss
```diff
@@ -0,0 +1,46 @@
...
+  font-size: $typography-font-size-base;
+}
+
+@mixin labelLDefaultBold() {
+  font-family: $typography-font-family;
+  font-weight: $typography-font-weight-bold;
+  line-height: $typography-line-height-2xs;
+  font-size: $typography-font-size-s;
+  letter-spacing: $typography-letter-spacing-s;
+  --token: label-l-default-bold;
```
kvas-damian, line 27: I'm not sure if it was your intention but custom attributes are inherited by child nodes.
We need to think about risks of this feature

lib/styles/settings/typography/_tokens.scss
```diff
@@ -0,0 +1,46 @@
...
+
+$typography-letter-spacing-s: var(--typography-letter-spacing-s);
+
+body {
+  font-family: $typography-font-family;
+  font-size: $typography-font-size-base;
+}
+
+@mixin labelLDefaultBold() {
+  font-family: $typography-font-family;
```
kvas-damian, line 22: I'm not convinced that we need `font-family` in tokens. Why do we need it? Is it in JSON file provided by UX?

lib/styles/settings/typography/_variables.scss
```diff
@@ -0,0 +1,16 @@
+:root {
+  --typography-font-family: 'Lato';
```
kvas-damian, line 2: if we really need this we should define full `font-family` string with fallback fonts. Right know it's possible that browser will render serif font while font is loading

lib/styles/settings/typography/_variables.scss
```diff
@@ -0,0 +1,16 @@
...
+
+  --typography-font-size-base: 16px;
+  --typography-font-size-s: 14px;
+  --typography-font-size-xs: 12px;
+  --typography-font-size-2xs: 10px;
+
+  --typography-font-weight-bold: bold;
```
kvas-damian, line 9: is it string or number in JSON?

lib/js/components/ProgressBar/ProgressBar.vue
```diff
@@ -196,21 +197,17 @@ $progress-bar-layers: (
 	}
 
 	&__labelText {
-		@include headlineXXS();
-		@include textBold;
+    font-size: var(--typography-font-size-xs);
```
kvas-damian, line 200: was it intended to not use token here? Is it real use-case or something?

lib/styles/settings/typography/_tokens.scss
damianmef: @kvas-damian there is one fix, but it increased code size:
```
--token: label-l-default-bold;
--token: initial;
```
Parent will have visible token (but with line-through) and childrens will have `--token: initial`.

lib/styles/settings/typography/_tokens.scss
damianmef: I think we can add condition for default font-family. If it will be our default `Lato` - we will not add it, otherwise we will add. So I'm removing it, it's redundant for me.

lib/styles/settings/typography/_variables.scss
damianmef: True. It's a such big think like change default font i whole css, so I think it can be manage by our styles, not from JSON.

lib/js/components/ProgressBar/ProgressBar.vue
damianmef: It's a real use-case. This line set smaller font size for mobile.

lib/styles/settings/typography/_tokens.scss
```diff
@@ -0,0 +1,46 @@
...
+  font-weight: $typography-font-weight-bold;
+  line-height: $typography-line-height-2xs;
+  font-size: $typography-font-size-s;
```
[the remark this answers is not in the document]
damianmef: Added.

lib/styles/settings/typography/_tokens.scss
damianmef: I think we should inform about that fact UX.

damianmef: Maybe this inheritance is not so stupid. Example:
1. Parent has `labelMDefaultBold();` with `--token: label-m-default-bold;` 
2. Children has no styles
3. Children of children inherits all parent styles and overwrites only font-size. So we will see in devtools:
a. Overwritten font size
b. all `labelMDefaultBold` rules with info where they come from - `--token: label-m-default-bold;`

kvas-damian: is 3. an advantage or disadvantage? What UX would like to see?

lib/styles/settings/typography/_tokens.scss
```diff
@@ -0,0 +1,58 @@
...
+@mixin labelLDefaultBold() {
+	--token: label-l-default-bold;
+
+	font-size: $typography-font-size-s;
+	font-style: $typography-font-style-normal;
+	font-weight: $typography-font-weight-bold;
+	letter-spacing: $typography-letter-spacing-s;
+	line-height: $typography-line-height-2xs;
+	text-decoration: $typography-text-decoration-none;
+	text-decoration: $typography-text-transform-none;
```
kvas-damian, line 33: wrong attribute name

lib/styles/settings/typography/_tokens.scss
kvas-damian: let's remember to communicate that to UX

lib/js/components/ProgressBar/ProgressBar.vue
kvas-damian: weh nwe introduce Typography tokens, do we plan to implement it this way? I thought, we'll be using only tokens, not variables...

lib/styles/settings/typography/_tokens.scss
kvas-damian: > Parent will have visible token (but with line-through) and childrens will have `--token: initial`.

I'm not sure if it's intuitive solution

lib/styles/settings/typography/_tokens.scss
damianmef: Fixed.

lib/js/components/ProgressBar/ProgressBar.vue
damianmef: Right, in FIgma there is another token for mobile. Fixed.

PR #218, opened 2022-10-31
IT-4808 - fix button animation fade in - in loading state

damianmef:


damianmef:
@kvas-damian after remove transition from content i was more visible that loading icon fade in animation is not working. I did many tries and only change this animation to keyframes fixed that. So now:
1. Content is disappears immediately.
2. Icon has 300ms fade in animation.
3. Loading -> default:
4. Icon disappears immediately.
5. Content shows immediately.

kvas-damian:
> and only change this animation to keyframes fixed that

Did you try usning animation name without typo: `a-fadeIn`? See definition in: https://github.com/bethinkpl/design-system/blob/987c93c6cd4d2ed3d400ebd6f3c7950e9045e28d/lib/styles/settings/_animations.scss#L11

lib/styles/components/_buttons.scss
```diff
@@ -101,6 +101,11 @@
 		&__loadingIcon {
 			color: $icon;
 		}
+
+		&__loadingIcon {
```
kvas-damian, line 105: why do we have to move it here?

damianmef: I like to have styles for element for constant classes, not for dynamic like `.loading`. Now it's not a problem, but sometimes it can occurs problems with animations. Dynamic class styles could be used for example: in changing opacity, from default 0 to 1 in loading state.

kvas-damian: makes sense, but do you think `setButtonAdditions()` mixin is the right place to put it?

damianmef: I moved to the right place.

lib/styles/components/_buttons.scss
```diff
@@ -336,4 +330,9 @@ $button-disabled-alpha: 0.6;
 			margin-left: $space-xxxxs;
 		}
 	}
+
+	&__loadingIcon {
+		position: absolute;
+		animation: $animation-fade-in, $animation-spin;
```
kvas-damian, line 336:
```suggestion
		animation: $animation-fade-in, $animation-spin;
		position: absolute;
```

kvas-damian (approved):

PR #220, opened 2022-11-02
IT-3400 - fix button ripple

damianmef:


lib/styles/components/_buttons.scss
```diff
@@ -168,6 +168,7 @@ $button-disabled-alpha: 0.6;
 		}
 	}
 
+	overflow: hidden;
```
kvas-damian, line 171: do we need it? If yes, should we introduce it in other component using `ripple`?

damianmef: I checked it by simulating resizing and probably not. But I will be sure when I will check it on the platform on the Watch button. So for now I'm removing this rule.

kvas-damian (approved):

PR #221, opened 2022-11-02
IT-4897 - working import and generating variables css/css

damianmef:


damianmef:
@kvas-damian I have two problems:
First:
I tried to split this file, but still cant import any my file. I know the source of the problem are our config files and any solution didn't work. I will back to this problem at the end of coding.
Second:
Do You have any low-cost solution how to convert JSON `fontWeight` node? Keys can have two properties like `bold italic` - so in loop there need to be some check -> split it and for every key make correct value. I mean: having key `bold italic` convert it to: `--typography-font-weight-bold: bold;` and `--typography-font-style-italic: italic;` and two separated SCSS variables. For me it's too complicated. These values and their combinations are very limited and are not our custom values like `font-size-3xl` so I think these variables can be hardcoded. They will not change, maybe we will add in the future `light italic` but now it's not worth to make this dynamic.

kvas-damian:
@damianmef is your comment still valid, or did you found a way to resolve these problems?

tools/importers/configs/SynchronizeTypographyTokensConfig.json
```diff
@@ -0,0 +1,19 @@
...
+        "variablesRaw": {
+          "destinationVariables": "_variables.scss",
+          "destinationVariablesCss": "_variables-css.scss",
+          "destinationJson": "_variables.json"
+        }
+      }
+    }
+  ],
+  "jsonBinApiUrl": "https://api.jsonbin.io/v3/b/",
+  "xMasterKey": "$2b$10$pGYu2VSnUSqm/6m2kMaJSOwc39KdqiTppeIv/qFcAaaqkRm.J/akW"
```
kvas-damian, line 18: it's a public repo. IMO, we need to regenerate this key now

tools/importers/SynchronizeTypographyTokens.ts
```diff
@@ -0,0 +1,138 @@
...
+
+	if (!response.data.record.values.LMSDesignSystemTypography) {
+		throw new TypeError('Response structure has no content!');
+	}
+
+	ImportTypographyRaw(
+		bin.name,
+		bin.files.variablesRaw,
+		response.data.record.values.LMSDesignSystemTypography,
+		bin.isTheme,
```
kvas-damian, line 128: it looks suspicious when parsing tokens, do we need to handle themes for typography?

tools/importers/SynchronizeTypographyTokens.ts
```diff
@@ -0,0 +1,138 @@
...
+
+	if (isTheme) {
+		resultCss.push('.theme-' + name + ' {');
+	} else {
+		resultCss.push(':root {');
+	}
+
+	for (let key in jsonTypography) {
+		if (!variablesExcludedKeys.includes(key)) {
+			let value = jsonTypography[key];
```
kvas-damian, line 39: what is this `value` represents? Can we make it more descriptive?

kvas-damian:
Superseded by https://github.com/bethinkpl/design-system/pull/222

PR #222, opened 2022-11-08
IT-4897 - Typography importer

damianmef:
… for colors and typography variables


damianmef:
@kvas-damian:
- I changed by the way colors stories from mdx to ts
- Pages for tokens are in progress
- I will update during the day this PR

damianmef:
@kvas-damian what kid of licence we have, MIT? Warning about missing type of licence was annoying, so I added in package.json:
```
"license": "UNLICENSED",
```

tools/importers/configs/SynchronizeTypographyTokensConfig.json
```diff
@@ -0,0 +1,23 @@
...
+        },
+        "tokens": {
+          "destination": "_tokens.scss",
+          "destinationJson": "_tokens.json"
+        }
+      }
+    }
+  ],
+  "jsonBinApiUrl": "https://api.jsonbin.io/v3/b/",
+  "xMasterKey": "$2b$10$pGYu2VSnUSqm/6m2kMaJSOwc39KdqiTppeIv/qFcAaaqkRm.J/akW"
```
damianmef, line 22: Will move it to .env

damianmef: Fixed.

kvas-damian: IMO, we have to rotate it. You can access git history i public repos

lib/js/styles/ItemsList.vue
```diff
@@ -0,0 +1,158 @@
...
+				</div>
+				<div
+					v-for="(items, itemCategory) in itemsList.list"
+					:key="itemCategory"
+					class="itemsList__row"
+				>
+					<h3 :id="'category-' + itemCategory">{{ itemCategory }}</h3>
+					<item-color v-if="type === TOKENS_TYPES.COLORS" :items="items" />
+
+					<!-- @todo replace content for ItemTypography -->
```
kvas-damian, line 25: don't forget

package.json
```diff
@@ -6,6 +6,7 @@
   "author": "nerdy@bethink.pl",
   "main": "./dist/design-system.umd.js",
   "typings": "./dist/index.d.ts",
+  "license": "UNLICENSED",
```
kvas-damian, line 9: did you contact Adam, to agree on license?

tools/importers/SynchronizeColorsTokens.ts
```diff
@@ -301,4 +247,4 @@ const SynchronizeColorsTokens = async () => {
 	});
 };
 
-SynchronizeColorsTokens();
+SynchronizeColorsTokens().then(() => console.log('Import in progress...'));
```
kvas-damian, line 250: something isn't right here. We wait until execution is done in this method, but we write "in progress"

tools/importers/helpers/modifiers.ts
```diff
@@ -0,0 +1,22 @@
+const camelize = (str: string) => {
+	return str
+		.replace(/(?:^\w|[A-Z]|\b\w)/g, function (word, index) {
+			return index === 0 ? word.toLowerCase() : word.toUpperCase();
+		})
+		.replace(/\s+/g, '');
+};
+
+const makeHexShortcut = (hex: string) => {
```
kvas-damian, line 9: it's color specific. IMO, it deserves a separate file

tools/importers/SynchronizeTypographyTokens.ts
```diff
@@ -0,0 +1,189 @@
...
+
+	if (!requestResponse.data.record.values.LMSDesignSystemTypography) {
+		throw new TypeError('Response structure has no content!');
+	}
+
+	return requestResponse;
+};
+
+const SynchronizeTypographyTokens = async () => {
+	tokensFilesConfig.bins.forEach((bin: ConfigFileBin) => {
```
kvas-damian, line 184: I'm not sure if keeping the same config structure as in Colors makes sense here. Now we have iteration over bins, but we know it's very unlikely we'll need to process two JSONs

tools/importers/helpers/structures.ts
```diff
@@ -0,0 +1,41 @@
...
+	label: string;
+	value: string;
+}
+
+export interface ConfigFileObject {
+	destinationVariables: string;
+	destinationVariablesCss: string;
+	destinationVariablesCssJson: string;
+	destinationJson: string;
+	destination: string;
```
kvas-damian, line 22: are they all required or optional? I see config file doesn't define all fields, so something is wrong

tools/importers/SynchronizeTypographyTokens.ts
```diff
@@ -0,0 +1,189 @@
...
+	name: string,
+	binValues: ConfigFileObject,
+	jsonTypography: Object,
+	isTheme: boolean,
+) => {
+	let resultCss: Array<string> = [];
+	let resultScss: Array<string> = [];
+	let resultJsonCss: Dict<Array<IResultJsonObject>> = {};
+
+	resultCss.push(modifiers.cssFileFirstLine(isTheme, name));
```
kvas-damian, line 29: we defined a `name` in config, we passed it through many layers and finally... It's ignored inside `cssFileFirstLine()`. Can we simplify it?

tools/importers/SynchronizeTypographyTokens.ts
```diff
@@ -0,0 +1,189 @@
...
+
+	for (let key in jsonTypography) {
+		if (!importerVariables.excludedKeys.includes(key)) {
+			let value: any = jsonTypography[key];
```
kvas-damian, line 33: we don't like `any` type...

kvas-damian, line 33: Can we find a variable name that is more descriptive? Maybe we could describe what is inside this `value`?

tools/importers/SynchronizeTypographyTokens.ts
```diff
@@ -0,0 +1,189 @@
...
+
+			for (let variableKey in value) {
+				let propertyName: string =
```
kvas-damian, line 36: `const` instead of `let`?

tools/importers/SynchronizeTypographyTokens.ts
```diff
@@ -0,0 +1,189 @@
...
+						const valueBase: number = parseInt(value.base.value.replace(/[^0-9]/g, ''));
+						const valueSplit: Array<number> = value[variableKey].value.split('*');
+						const valueMultiplier: number = valueSplit[1];
+						propertyValue = valueMultiplier
+							? valueBase * valueMultiplier + 'px'
+							: valueBase + 'px';
+						break;
+					case importerVariables.tokensTypographyInputAttributes.letterSpacing:
+						propertyValue =
+							value[variableKey].value.replace(/[^0-9/.]/g, '') * 0.01 + '';
```
kvas-damian, line 55: why do we concatenate it with empty string?

tools/importers/SynchronizeTypographyTokens.ts
```diff
@@ -0,0 +1,189 @@
...
+		tokensFilesConfig.destinationPath + binValues.destinationVariablesCssJson,
+		resultJsonCss,
+	);
+};
+
+const ImportTypographyTokensRaw = (binValues: ConfigFileObject, jsonTypography: any) => {
+	let resultScss: Array<ITypographyToken> = [];
+
+	for (let key in jsonTypography[importerVariables.tokensKey]) {
+		let token: ITypographyToken = fileRead.recursiveTokenReader(
```
kvas-damian, line 109: why is it called `fileRead(er)`? It isn't reading a file.

tools/importers/helpers/fileReader.ts
```diff
@@ -0,0 +1,59 @@
+const modifiers = require('./modifiers');
+
+const recursiveTokenReader = (obj, keyResult: string, result: Array<any>, importerVariables) => {
```
kvas-damian, line 3: do we need to pass `result` as a parameter? Can't we merge results at the end of this function?

tools/importers/SynchronizeTypographyTokens.ts
```diff
@@ -0,0 +1,189 @@
...
+				cssAttributePartOne +
+					'-' +
+					cssAttributePartTwo +
+					': $' +
+					importerVariables.typographyPrefix +
+					attribute +
+					';',
+			);
+		});
+		result.push('}');
```
kvas-damian, line 141: I believe token name is missing

lib/styles/settings/typography/_variables-css.json
```diff
@@ -0,0 +1 @@
...defined_typography-font-size-2xs","label":"typography-font-size-2xs","value":"10px"},{"id":"undefined_typography-font-size-3xs","label":"typography-font-size-3xs","value":"8px"}],"letterSpacing":[{"id":"undefined_typography-letter-spacing-xs","label":"typography-letter-spacing-xs","value":"0"},{"id":"undefined_typography-letter-spacing-s","label":"typography-letter-spacing-s","value":"0.005em"},{"id":"undefined_typography-letter-spacing-m","label":"typography-letter-spacing-m","value":"0.0075em"},{"id":"undefined_typography-letter-spacing-l","label":"typography-letter-spacing-l","value":"0.015em"},{"id":"undefined_typography-letter-spacing-xl","label":"typography-letter-spacing-xl","value":"0.06em"}],"textCase":[{"id":"undefined_typography-text-transform-none","label":"typography-text-transform-none","value":"none"},{"id":"undefined_typography-text-transform-uppercase","label":"typography-text-transform-uppercase","value":"uppercase"}],"textDecoration":[{"id":"undefined_typography-text-decoration-none","label":"typography-text-decoration-none","value":"none"},{"id":"undefined_typography-text-decoration-underline","label":"typography-text-decoration-underline","value":"underline"}]}
```
kvas-damian, line 1: is undefined ok in this file?

tools/importers/helpers/typographyVariables.ts
```diff
@@ -0,0 +1,53 @@
...
+	fontWeight: 'font-weight',
+	fontSize: 'font-size',
+	letterSpacing: 'letter-spacing',
+	textCase: 'text-transform',
+	textDecoration: 'text-decoration',
+};
+const resultCssAdditionalLines: Array<string> = [
+	'--typography-font-weight-normal: normal;',
+	'--typography-font-weight-bold: bold;',
+	'--typography-font-weight-light: lighter;',
```
kvas-damian, line 22: https://htmldog.com/references/css/properties/font-weight/#:~:text=lighter,inherited%20font%20weight

tools/importers/configs/SynchronizeTypographyTokensConfig.json
damianmef: I changed the key.

tools/importers/helpers/modifiers.ts
damianmef: Done.

tools/importers/SynchronizeTypographyTokens.ts
damianmef: I splitted config interface.

tools/importers/SynchronizeTypographyTokens.ts
```diff
@@ -0,0 +1,189 @@
...
+}
+
+const SynchronizeSingleBin = async (bin: ConfigFileBin) => {
+	let requestResponse = await requestForBin(bin);
+
+	ImportTypographyRaw(
+		bin.name,
+		bin.files.variablesRaw,
+		requestResponse.data.record.values.LMSDesignSystemTypography,
+		bin.isTheme,
```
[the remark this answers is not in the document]
damianmef: Fixed.

tools/importers/helpers/structures.ts
damianmef: Fixed.

tools/importers/SynchronizeTypographyTokens.ts
damianmef: Fixed.

tools/importers/SynchronizeTypographyTokens.ts
damianmef: Fixed.

damianmef: Updated, please check.

tools/importers/SynchronizeTypographyTokens.ts
damianmef: Fixed.

tools/importers/SynchronizeTypographyTokens.ts
damianmef: After adding typing to whole `value` i need to split this operations.

tools/importers/helpers/typographyVariables.ts
```diff
@@ -0,0 +1,53 @@
...
+	'text-case': 'text-transform',
+	'-regular': '-normal',
+};
+const importVariables: string = "@import 'variables';";
+const jsonAttributeToCssProperty: { [key: string]: string } = {
+	lineHeight: 'line-height',
+	fontWeight: 'font-weight',
+	fontSize: 'font-size',
+	letterSpacing: 'letter-spacing',
+	textCase: 'text-transform',
```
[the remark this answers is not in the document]
damianmef: Yes, updated.

tools/importers/SynchronizeTypographyTokens.ts
damianmef: Because I'm reading the file content there. Maybe `fileCrawler` @kvas-damian ?

tools/importers/SynchronizeTypographyTokens.ts
damianmef: Good catch, fixed.

lib/styles/settings/typography/_variables-css.json
damianmef: Fixed.

tools/importers/helpers/typographyVariables.ts
damianmef: I changed it to numbers.

tools/importers/SynchronizeColorsTokens.ts
damianmef: I can left empty body here, but because we have async requests it always is displaying first, so in the practice we have to lines in output:
```
Import in progress...
The import was successful for bin: 633d96230e6a79321e1dec41
```

tools/importers/configs/SynchronizeTypographyTokensConfig.json
kvas-damian: does UX use the same key? Do we need to inform them that the old one isn't valid anymore?

tools/importers/SynchronizeColorsTokens.ts
kvas-damian: you can `await` inside the method to mitigate this problem. Also, if you decide to not change that, we shouldn't be returning Promise - we're not waiting for execution to be finished, so it's misleading

tools/importers/SynchronizeTypographyTokens.ts
kvas-damian: we're not touching file. We operate on a tree of variables. I wouldn't call it `fileAnything`

package.json
damianmef: `UNLICENSED` is ok.

tools/importers/helpers/fileReader.ts
```diff
@@ -0,0 +1,65 @@
...
+	transformCssProperty,
+} = require('../helpers/typographyVariables');
+
+export const recursiveTokenReader = (obj, keyResult: string, result: Array<ITypographyToken>) => {
+	if ('value' in obj) {
+		return;
+	}
+
+	for (let key in obj) {
+		let temporaryKey = (keyResult + ' ' + key).replace(/\-+/g, ' ');
```
kvas-damian, line 17: don't we need to skip `deafult` in token name?

lib/js/styles/ItemsList.vue
damianmef: I will touch it again later, don't close this thread.

tools/importers/helpers/fileReader.ts
damianmef: Fixed, parameter removed.

tools/importers/SynchronizeColorsTokens.ts
damianmef: I fixed it, now we have information in correct order and time:
```
Import in progress...
The import was successful for bin: 633d96230e6a79321e1dec41
The importer has completed the work.
```

tools/importers/SynchronizeTypographyTokens.ts
damianmef: Renamed filed to `tokensReader.ts`

tools/importers/helpers/fileReader.ts
damianmef: We need, fixed.

tools/importers/configs/SynchronizeTypographyTokensConfig.json
damianmef: @kvas-damian problem solved. After consultation with UX I created 2 separated `X-ACCESS-KEY`:
1. Figma - with full permissions - UX is testing it in Figma.
2. Importers - only for reading.
So we are not using `X-MASTER-KEY`, I also updated passwd.

tools/importers/SynchronizeColorsTokens.ts
kvas-damian: it's good that order of logs is correct, but I still believe that if we return a `Promise` it should be possible to await until code in `Promise` is executed. Right now, you attach to `then()` which is fired when `Promise` is resolved, but you console log is printed before code inside this method finishes

tools/importers/SynchronizeTypographyTokens.ts
```diff
@@ -0,0 +1,191 @@
...
+} from './helpers/structures';
+
+dotenv.config();
+
+const ImportTypographyVariables = (binFilesConfig: TypographyBinFiles, jsonTypography: Object) => {
+	let resultCss: Array<string> = [];
+	let resultScss: Array<string> = [];
+	let resultJsonCss: Dict<Array<IResultJsonObject>> = {};
+
+	resultCss.push(cssFileFirstLine(false, undefined));
```
kvas-damian, line 40: maybe we should have a separate method for theme and for `:root` css block start?

tools/importers/SynchronizeTypographyTokens.ts
```diff
@@ -0,0 +1,191 @@
...
+						propertyValue = valueMultiplier
+							? valueBase * valueMultiplier + 'px'
+							: valueBase + 'px';
+						break;
+					case tokensTypographyInputAttributes.letterSpacing:
+						const temporaryValueAsString: string = propertyOptions[
+							variableKey
+						].value.replace(/[^0-9/.]/g, '');
+						const temporaryValueAsNumber: number = +temporaryValueAsString * 0.01;
+						propertyValue = temporaryValueAsNumber + '';
```
kvas-damian, line 69: https://peps.python.org/pep-0020/#:~:text=Explicit%20is%20better%20than%20implicit.
> Explicit is better than implicit.

```suggestion
						propertyValue = temporaryValueAsNumber.toString();
```

tools/importers/helpers/tokensReader.ts
```diff
@@ -0,0 +1,64 @@
...
+				tokenCamelCase: camelize(keyResult),
+				attributes,
+			},
+		];
+	}
+
+	let results: Array<ITypographyToken> = [];
+	for (let key in obj) {
+		let newSuffix: string = !key.includes(tokenPartDisabled) ? key : '';
+		let temporaryKey: string = (keyResult + ' ' + newSuffix).replace(/\-+/g, ' ');
```
kvas-damian, line 58: if `keyResult` is an empty string, won't we get double space here? 
I understand that you use regex to mitigate this problem, but if someone else starts using this code, he could find it difficult to understand what is going on

package.json
```diff
@@ -68,7 +71,7 @@
     "stylelint-scss": "3.18.0",
     "ts-jest": "^26.5.6",
     "ts-node": "^10.9.1",
-    "typescript": "4.2.4",
+    "typescript": "4.8.4",
```
kvas-damian, line 74: did you check if production build (`yarn build`) still works after this change?

tools/importers/SynchronizeTypographyTokens.ts
```diff
@@ -0,0 +1,191 @@
...
+	);
+
+	if (!requestResponse.data.record.values.LMSDesignSystemTypography) {
+		throw new TypeError('Response structure has no content!');
+	}
+
+	return requestResponse;
+};
+
+SynchronizeTypographyTokensBin().finally(() => {
```
kvas-damian, line 189: there is a lot of logic in this importer. Maybe we should cover it with tests

tools/importers/SynchronizeTypographyTokens.ts
damianmef: Fixed.

tools/importers/SynchronizeTypographyTokens.ts
damianmef: Sure, splitted.

tools/importers/helpers/tokensReader.ts
damianmef: Fixed, I removed by the way `trim()` and need change `camelize` to `pascalCase` for creating mixin name.

package.json
damianmef: Hmm right, there is the error. Generally solution everywhere is to bump to latest version `ts-node` and `typescript` but in this case error occurs. I downgraded version of `typescript` to `4.6.4` and it works fine. It doesn't make much difference to us, we'll wait until they fix the latest versions of both libraries.

tools/importers/SynchronizeColorsTokens.ts
damianmef: Wait, I made change in the tokens importer, not here. I fixed it and now it looks like this:
```
SynchronizeColorsTokens().finally(() => {
	console.log('The importer has completed the work.');
});
```

damianmef: Ok, I think console log there is redundant in this case. I removed it. Now we have info about:
```
Import in progress...
The import was successful for bin: 633d96230e6a79321e1dec41
```

tools/importers/SynchronizeTypographyTokens.ts
damianmef: Testes for typography synchronization added. Please check used convention. I will make tests for colors synchronization only if these will be acceptable.

tools/importers/SynchronizeTypographyTokens.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+	ImportTypographyVariables,
+	ImportTypographyTokens,
+	requestForBin,
+} from './SynchronizeTypographyTokens';
+const fs = require('fs');
+const tokensFilesConfig = require('./configs/FakeSynchronizeTypographyTokensConfig.json');
+
+describe('Typography Tokens', () => {
+	it('check config & create files', async () => {
+		let requestResponse = await requestForBin(
```
kvas-damian, line 11: do we request external service in tests?
It isn't a good practice. What if UX updates JSON file, so we get different items than we expect?

tools/importers/SynchronizeTypographyTokens.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+			requestResponse.data.record.values.LMSDesignSystemTypography,
+		);
+
+		ImportTypographyTokens(
+			tokensFilesConfig,
+			requestResponse.data.record.values.LMSDesignSystemTypography,
+		);
+	});
+	it('check variable SCSS file', async () => {
+		await new Promise((r) => setTimeout(r, 500));
```
kvas-damian, line 27: sleep in tests code? We should avoid such patterns. What if operation takes longer than `500ms`?

tools/importers/SynchronizeTypographyTokens.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+		let variableFileRaw = await fs.promises.readFile(
```
kvas-damian, line 28: does this test depends on previous test? What if we start running tests in parallel?

tools/importers/SynchronizeTypographyTokens.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+	it('check tokens JSON file', async () => {
+		let variableFileJsonTokens = await fs.promises.readFile(
+			tokensFilesConfig.destinationPath + tokensFilesConfig.bin.files.tokens.destinationJson,
+			'utf8',
+		);
+		let variableFileSplit = JSON.parse(variableFileJsonTokens);
+		expect(variableFileSplit.DisplayHeading[0].token).toBe('display-heading-m-regular');
+		expect(variableFileSplit.DisplayHeading[0].attributes).toContain('font-size-4xl');
+	});
+	it('unlink temporary files', async () => {
```
kvas-damian, line 80: it would be great if we could mock/spy filesystem. Then you won't need to remove files. 
Additionally now, if something fails, we end up with random files

lib/js/styles/Colors/Colors.stories.ts
```diff
@@ -0,0 +1,26 @@
+import { Args, Meta, StoryFn } from '@storybook/vue';
+
+import ItemsList from './../ItemsList.vue';
+import source from '../../../styles/settings/colors/_raw.json';
```
kvas-damian, line 4: what `source` means here?

kvas-damian: the same question for other files

lib/js/styles/ColorsTokensLms/ColorsTokensLms.stories.ts
```diff
@@ -0,0 +1,26 @@
+import { Args, Meta, StoryFn } from '@storybook/vue';
+
+import ItemsList from './../ItemsList.vue';
+import source from '../../../styles/settings/colors/_tokens.json';
+import { TOKENS_TYPES } from '../TokenTypes.consts';
+
+export default {
+	title: 'foundations/Design Tokens/Color Tokens/ LMS',
```
kvas-damian, line 8:
```suggestion
	title: 'foundations/Design Tokens/Color Tokens/LMS',
```

lib/js/styles/ItemColor.vue
```diff
@@ -0,0 +1,27 @@
...
+<style lang="scss" scoped>
+@import '../../styles/components/tokens';
+</style>
+
+<script lang="ts">
+export default {
+	name: 'ItemColor',
+	props: {
+		items: {
+			type: Array,
```
kvas-damian, line 22: maybe we could define type of this prop while we're touching this file?

lib/js/styles/ItemTypography.vue
```diff
@@ -0,0 +1,55 @@
...
+<style lang="scss" scoped>
+@import '../../styles/components/tokens';
+</style>
+
+<script lang="ts">
+export default {
+	name: 'ItemTypography',
+	props: {
+		items: {
+			type: Array,
```
kvas-damian, line 38: this is a complex type. What do you think about adding `PropType` to make it easier to uderstand what we should pass into this component?

lib/js/styles/ItemTypography.vue
```diff
@@ -0,0 +1,55 @@
...
+		},
+	},
+	methods: {
+		attributesAsStyle(attributes) {
+			let result: string = '';
+			attributes.forEach((item) => {
+				result += item.property + ': var(--typography-' + item.value + ');';
+			});
+
+			console.log(result);
```
kvas-damian, line 49: debugging code?

lib/js/styles/ItemTypography.vue
```diff
@@ -0,0 +1,55 @@
...
+		<div v-for="item in items" :key="item.id" class="itemDefinition">
+			<div v-if="item.label" class="itemDefinition__label">{{ item.label }}</div>
+			<div v-if="item.value" class="itemDefinition__value">{{ item.value }}</div>
+			<div v-if="item.token" class="itemDefinition__token">{{ item.token }}</div>
+			<div v-if="item.attributesRaw" class="itemDefinition__attributes">
+				<div v-for="attribute in item.attributesRaw" :key="item.id + attribute.value">
+					<strong>{{ attribute.property }}</strong
+					>:
+					<span>
+						<span v-pre>{</span>
```
kvas-damian, line 13: can we use `template` tag to limit number of DOM nodes?

lib/js/styles/ItemsList.vue
```diff
@@ -0,0 +1,141 @@
...
+			TOKENS_TYPES: Object.freeze(TOKENS_TYPES),
+		};
+	},
+	mounted() {
+		this.itemsLocal.forEach((list, index) => {
+			if (list.disabled) {
+				for (let key in list.list) {
+					if (key === list.disabled) {
+						this.itemsLocal[index].list[key] = null;
+						delete this.itemsLocal[index].list[key];
```
kvas-damian, line 134: Vue doesn't like when we modify props inside component: https://stackoverflow.com/a/43701284

lib/js/styles/ItemsList.vue
```diff
@@ -0,0 +1,141 @@
...
+			itemsLocal: this.itemsLists,
+			TOKENS_TYPES: Object.freeze(TOKENS_TYPES),
+		};
+	},
+	mounted() {
+		this.itemsLocal.forEach((list, index) => {
+			if (list.disabled) {
+				for (let key in list.list) {
+					if (key === list.disabled) {
+						this.itemsLocal[index].list[key] = null;
```
kvas-damian, line 133: do we need to set it to `null` and delete?

lib/js/styles/ItemsList.vue
```diff
@@ -0,0 +1,141 @@
...
+		},
+	},
+	data() {
+		return {
+			itemsLocal: this.itemsLists,
+			TOKENS_TYPES: Object.freeze(TOKENS_TYPES),
+		};
+	},
+	mounted() {
+		this.itemsLocal.forEach((list, index) => {
```
kvas-damian, line 129: isn't it better to use `computed`?

lib/js/styles/ItemsList.vue
```diff
@@ -0,0 +1,141 @@
+<template>
+	<div>
+		<div v-for="itemsList in itemsLocal" :key="itemsList.title" class="itemDefinition">
+			<div class="itemsList" :class="{ [itemsList.class]: itemsList.class }">
```
kvas-damian, line 4:
```suggestion
			<div class="itemsList" :class="itemsList.class">
```

lib/js/styles/ItemColor.vue
```diff
@@ -0,0 +1,27 @@
...
+	</div>
+</template>
+
+<style lang="scss" scoped>
+@import '../../styles/components/tokens';
+</style>
+
+<script lang="ts">
+export default {
+	name: 'ItemColor',
```
kvas-damian, line 19: Is it an `Item` or list of `Items`?

lib/js/styles/ItemsList.vue
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	&__category {
+		font-weight: var(--typography-font-weight-bold);
+		font-size: var(--typography-font-size-2xl);
+		line-height: var(--typography-line-height-3xl);
+	}
+
+	&__row {
+		border: 1px solid #f3f2f2;
```
kvas-damian, line 49: is it possible to use css token varibale for color?

lib/js/styles/ItemsList.vue
```diff
@@ -0,0 +1,141 @@
...
+		margin-bottom: 20px;
+		padding: 20px 20px;
```
kvas-damian, line 51: maybe we should try to import DS variables for styling DS? ;)

lib/js/styles/ItemTypography.vue
```diff
@@ -0,0 +1,55 @@
+<template>
+	<div>
+		<!-- @todo fill component -->
```
kvas-damian, line 3: TODO?

lib/js/styles/ItemsList.vue
```diff
@@ -0,0 +1,141 @@
+<template>
+	<div>
+		<div v-for="itemsList in itemsLocal" :key="itemsList.title" class="itemDefinition">
```
kvas-damian, line 3: should this component use the same class name as child components?

lib/js/styles/ItemsList.vue
```diff
@@ -0,0 +1,141 @@
...
+	}
+}
+
+.itemDefinition {
+	align-items: center;
+	display: flex;
+	justify-content: space-between;
+	width: 100%;
+
+	&__id,
```
kvas-damian, line 65: do we use these styles?

lib/styles/components/_tokens.scss
```diff
@@ -0,0 +1,66 @@
...
+			font-weight: var(--typography-font-weight-bold);
+		}
+	}
+
+	&__id {
+		flex: 1 1 auto;
+		width: 200px;
+	}
+
+	&__quarter {
```
kvas-damian, line 58: not used?

lib/styles/components/_tokens.scss
```diff
@@ -0,0 +1,66 @@
...
+
+	&__tile {
+		box-shadow: rgb(0 0 0 / 10%) 0 1px 3px 0;
+		display: block;
+		height: 40px;
+		width: 100%;
+	}
+
+	&__value {
+		color: #707070;
```
kvas-damian, line 28: can we use token?

lib/styles/settings/colors/_raw-bodywork.json
```diff
@@ -1 +1 @@
-{"default":[{"id":"_raw-bodywork.scss_white","label":"white","value":"#FFF"}],"theme":[{"id":"_raw-bodywork.scss_theme-50","label":"theme-50","value":"#ECF2F5"},{"id":"_raw-bodywork.scss_theme-100","label":"theme-100","value":"#DCE7ED"},{"id":"_raw-bodywork.scss_theme-200","label":"theme-200","value":"#B6CDD7"},{"id":"_raw-bodywork.scss_theme-300","label":"theme-300","value":"#81B2C8"},{"id":"_raw-bodywork.scss_theme-400","label":"theme-400","value":"#3D8BAF"},{"id":"_raw-bodywork.scss_theme-500","label":"theme-500","value":"#1F7095"},{"id":"_raw-bodywork.scss_theme-600","label":"theme-600","value":"#0E5D81"},{"id":"_raw-bodywork.scss_theme-700","label":"theme-700","value":"#004C6E"},{"id":"_raw-bodywork.scss_theme-800","label":"theme-800","value":"#0A425B"},{"id":"_raw-bodywork.scss_theme-900","label":"theme-900","value":"#183340"}]}
```
kvas-damian, line 1: was it intended that we changed `id` for raw colors in this PR?

lib/js/styles/ItemColor.vue
```diff
@@ -0,0 +1,27 @@
...
+			<div class="itemDefinition__value">{{ item.value }}</div>
+			<div class="itemDefinition__color">
+				<span :style="{ background: item.value }" class="itemDefinition__tile" />
+			</div>
+		</div>
+	</div>
+</template>
+
+<style lang="scss" scoped>
+@import '../../styles/components/tokens';
```
kvas-damian, line 14: are you sure, we're importing tokens here?

tools/importers/SynchronizeTypographyTokens.ts
```diff
@@ -173,29 +187,33 @@ function buildTypographyTokensMixins(tokens: Array<ITypographyToken>): Array<str
...
-		tokensFilesConfig.bin.files,
+		tokensFilesConfig,
 		requestResponse.data.record.values.LMSDesignSystemTypography,
 	);
 
 	console.log('The import was successful for bin: ' + tokensFilesConfig.bin.id);
 };
 
-const requestForBin = async (binConfig: TypographyConfigFileBin) => {
+export const requestForBin = async (binConfig: TypographyConfigFileBin, jsonBinApiUrl: string) => {
```
kvas-damian, line 209: are you sure that we should export this function? Does it makes sense for devs to call it separately from the rest of the code?

tools/importers/SynchronizeTypographyTokens.spec.ts
```diff
@@ -0,0 +1,80 @@
+import {
+	ImportTypographyVariables,
+	ImportTypographyTokens,
+	requestForBin,
+} from './SynchronizeTypographyTokens';
+const fs = require('fs');
+const tokensFilesConfig = require('./configs/FakeSynchronizeTypographyTokensConfig.json');
```
kvas-damian, line 7: do we need to keep config in separate file? Can't we define it in test file?

tools/importers/SynchronizeColorsTokens.ts
kvas-damian: IMO, the root cause of the problem is here: https://github.com/bethinkpl/design-system/blob/IT-4898-tt-tokens-to-scss/tools/importers/SynchronizeColorsTokens.ts#L254

You define function as `async`, but inside the function you don't `await` for asynchronous functions execution. Because of that code in `then/finally` is executed before asynchronous code (`axios.get()`) is finished

tools/importers/SynchronizeTypographyTokens.spec.ts
damianmef: I know it's not the best practice. But if something will be broken in the JSON files from Figma.

> What if UX updates JSON file, so we get different items than we expect?

Hmm we should know about it, but it true it shouldn't break CI at a random moment. I will convert it to local files and we will get rid of other problems.

tools/importers/SynchronizeTypographyTokens.spec.ts
damianmef: I don't know where is the problem, without this timeout tests sometimes passed sometimes not. They are launched sequentially but maybe when file is written on the disc in the first test, in the second is not available yet. But if I will change request to the static file, this problem will not occurs.

tools/importers/SynchronizeTypographyTokens.spec.ts
damianmef: Same answer as here: https://github.com/bethinkpl/design-system/pull/222#discussion_r1023903215

tools/importers/SynchronizeTypographyTokens.spec.ts
damianmef: Ok.

lib/js/styles/Colors/Colors.stories.ts
damianmef: It means `inputData` have You better name than `listData`?

lib/js/styles/ItemTypography.vue
damianmef: I will try.

lib/js/styles/ItemColor.vue
damianmef: Ok.

lib/styles/settings/colors/_raw-bodywork.json
damianmef: I changed keys in config files for Typography and Colors. It's only used in for loop in view to have different ID for every item. It has no other consequences.

tools/importers/SynchronizeColorsTokens.ts
```diff
@@ -296,9 +252,10 @@ const SynchronizeSingleBin = async (bin) => {
...
 const SynchronizeColorsTokens = async () => {
-	tokensFilesConfig.bins.forEach((bin) => {
+	console.log('Import in progress...');
+	tokensFilesConfig.bins.forEach((bin: ColorsConfigFileBin) => {
 		SynchronizeSingleBin(bin);
 	});
 };
 
-SynchronizeColorsTokens();
+SynchronizeColorsTokens().finally(() => {});
```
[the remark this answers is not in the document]
damianmef: No, but without any action after `SynchronizeColorsTokens()` IDE shows warning and preffer add `then()` with message `Promise returned from SynchronizeTypographyTokensBin is ignored `.

lib/styles/settings/colors/_raw-bodywork.json
kvas-damian: did you see that now it's called `token`?
Also in token file, then `id` starts with `raw` now...

tools/importers/SynchronizeTypographyTokens.spec.ts
damianmef: It is needed now in testes, but if I will remove it from tests, this export will be redundant.

lib/js/styles/Colors/Colors.stories.ts
kvas-damian: `rawColorsList`?

tools/importers/SynchronizeTypographyTokens.spec.ts
damianmef: > it isn't good practice to change API for tests.
> Tests should try to use code in the same way as we do in production code.

I agree, but if code was written completely without thinking how to use it in tests, or it was not split enought - it should be changed.

lib/js/styles/ItemsList.vue
damianmef: Removed.

tools/importers/SynchronizeTypographyTokens.ts
damianmef: It's reverted.

lib/js/styles/ItemColor.vue
damianmef: Yes, because these are styles for visual component tokens (colors, typography) not tokens as variablels.

lib/styles/settings/colors/_raw-bodywork.json
damianmef: Right, fixed.

lib/styles/components/_tokens.scss
damianmef: Fixed.

lib/styles/components/_tokens.scss
damianmef: Yes, removed.

lib/js/styles/ItemsList.vue
damianmef: No, removed.

lib/js/styles/ItemsList.vue
damianmef: Cleaned up.

lib/js/styles/ItemTypography.vue
damianmef: Removed.

lib/js/styles/ItemsList.vue
damianmef: I changed it, but we have no values for `20px` in spacings. Considering typography where UX didn't used the new typography scale, here I think they will not check if it is `20px`, I used close values in whole file, here for example `24px`.

lib/js/styles/ItemsList.vue
damianmef: Fixed.

lib/js/styles/ItemColor.vue
damianmef: Fixed.

lib/js/styles/ItemsList.vue
damianmef: Can be.

lib/js/styles/ItemsList.vue
damianmef: No, fixed.

lib/js/styles/ItemTypography.vue
damianmef: Yes, fixed.

lib/js/styles/ItemTypography.vue
damianmef: Removed.

lib/js/styles/ItemTypography.vue
damianmef: Updated.

lib/js/styles/ItemColor.vue
damianmef: Updated.

lib/js/styles/ColorsTokensLms/ColorsTokensLms.stories.ts
damianmef: Fixed.

lib/js/styles/Colors/Colors.stories.ts
damianmef: It's fine, updated.

tools/importers/SynchronizeTypographyTokens.spec.ts
damianmef: Are You sure about mocked data? I think the best test is where we have real data generated by importer. Now I'm using only mocked input data, the rest is based on dynamic data.

tools/importers/SynchronizeTypographyTokens.spec.ts
damianmef: This problem still occurs, I mean when I run importer and immediately checking files it always fails. But when I added:
```
await new Promise((r) => setTimeout(r, 1));
```
it's fine. To avoid dependency from tests I put everything in the one test. I couldn't read the file before tests, because only single test can be async.

damianmef: And I stayed with the solution create -> check -> delete files.

tools/importers/SynchronizeTypographyTokens.spec.ts
damianmef: https://github.com/bethinkpl/design-system/pull/222#discussion_r1024808141

tools/importers/SynchronizeTypographyTokens.spec.ts
damianmef: Fixed, we are using now static file, which has minimum structure.

tools/importers/SynchronizeColorsTokens.ts
damianmef: Right, but if function is not `async`  the problem is only moved on -> we are doing nothing with async `SynchronizeSingleBin()` and IDE has the same problem. Generally we need somewhere catch the promise.

lib/js/styles/ItemTypography.vue
damianmef: It can't be used, `<template>` need have directive and `v-pre` is not acceptable by lint.

damianmef: I just added a space between brackets:
```
<span>{ {{ attribute.value }} }</span>
```
I think it will be not a problem for UX.

lib/js/styles/ItemColor.vue
kvas-damian: It's confusing that scss file name and component name that are related, don't have anything in common. Do you think we should fix it?

lib/js/styles/ItemsList.vue
```diff
@@ -125,17 +90,20 @@ export default {
...
-						this.itemsLocal[index].list[key] = null;
-						delete this.itemsLocal[index].list[key];
-					}
-				}
-			}
-		});
-	},
+  computed: {
+    getItemsLocal() {
+      this.itemsLocal.forEach((list, index) => {
```
kvas-damian: I though we'll use `itemsLists` prop in the computed - create a local copy and then perform modifications.

Right now, if you change prop value, component won't be updated

lib/js/styles/ItemsColor.vue
```diff
@@ -0,0 +1,29 @@
...
+</style>
+
+<script lang="ts">
+import { ColorToken } from './TokenTypes';
+
+export default {
+	name: 'ItemsColor',
+	props: {
+		items: {
+			type: Array as () => Array<ColorToken>,
```
kvas-damian, line 24:
```suggestion
			type: Array PropType<Array<ColorToken>>,
```

lib/js/styles/ItemsList.vue
```diff
@@ -0,0 +1,109 @@
...
+
+export default {
+	name: 'ItemsList',
+	components: {
+		ItemsColor,
+		ItemsTypography,
+	},
+	props: {
+		itemsLists: {
+			type: Array,
```
kvas-damian, line 76: type?

lib/js/styles/ItemsTypography.vue
```diff
@@ -0,0 +1,50 @@
...
+</style>
+
+<script lang="ts">
+import { TypographyToken } from './TokenTypes';
+
+export default {
+	name: 'ItemsTypography',
+	props: {
+		items: {
+			type: Array as () => Array<TypographyToken>,
```
kvas-damian, line 35:
```suggestion
			type: Array as PropType<Array<TypographyToken>>,
```

lib/styles/components/_tokens.scss
kvas-damian: really?

tools/importers/.cache/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Ignore everything in this directory
+*
```
kvas-damian, line 2: is this file still in use?

tools/importers/SynchronizeTypographyTokens.spec.ts
```diff
@@ -0,0 +1,92 @@
+import { ImportTypographyVariables, ImportTypographyTokens } from './SynchronizeTypographyTokens';
+const fs = require('fs');
+const tokensFilesConfig = require('./configs/FakeSynchronizeTypographyTokensConfig.json');
```
kvas-damian, line 3: instead of creating new file, can we define this valiable this file? This file could be confusing when you aren't aware it's only used in tests

tools/importers/SynchronizeTypographyTokens.spec.ts
kvas-damian: did you try mocking `fs` instead of reading a file?

tools/importers/helpers/fileWriter.ts
```diff
@@ -0,0 +1,50 @@
+import { WriteStream } from 'fs';
+
+const fs = require('fs');
```
kvas-damian, line 3: why not `import`?

.github/workflows/ci.yml
```diff
@@ -24,6 +24,10 @@ jobs:
         run: yarn install --pure-lockfile
       - name: TypeScript check
         run: yarn ts:check
+      - name: Make envfile
+        uses: SpicyPizza/create-envfile@v1.3
+        with:
+          envkey_JSON_BIN_X_MASTER_KEY: ${{ secrets.JSON_BIN_X_MASTER_KEY }}
```
kvas-damian, line 30: is it still required?

lib/styles/settings/typography/_tokens.json
```diff
@@ -0,0 +1,12245 @@
+{
+  "DisplayHeading": [
```
kvas-damian, line 2: does our parser prettify json output, or will I get this file updated when I synchronize token again?

damianmef: It was manual reformat by me, I reverted it.

tools/importers/helpers/fileWriter.ts
damianmef: I refactored reuqire's to imports.

.github/workflows/ci.yml
damianmef: No, I removed it also from secrets on the github. Also renamed it variable in code to `JSON_BIN_X_ACCESS_KEY`.

tools/importers/.cache/.gitignore
damianmef: No, removed.

tools/importers/SynchronizeTypographyTokens.spec.ts
damianmef: I'm not sure if I understand you correctly. I removed file and move it's content to variable in the spec file.

lib/js/styles/ItemsTypography.vue
damianmef: Fixed.

lib/js/styles/ItemsList.vue
damianmef: Fixed.

lib/js/styles/ItemsColor.vue
damianmef: Fixed.

lib/js/styles/ItemColor.vue
damianmef: If you think that's confusing, sure. Changed.

tools/importers/SynchronizeTypographyTokens.spec.ts
damianmef: > did you try mocking fs of fileWriter instead of reading a file?

I don't see the point of mocking here. At the start I run importer and it generates files with mocked input JSON. If I will mock results of importer, these tests only could check if importer was not crashed.

lib/js/styles/ItemsList.vue
damianmef: Fixed.

lib/styles/components/_tokens.scss
damianmef: hmm, now yes

lib/js/styles/ItemsList.vue
```diff
@@ -31,32 +31,7 @@
...
-		border: 1px solid var(--raw-gray-100);
-		margin-bottom: $space-m;
-		padding: $space-m;
-	}
-
-	&__categories {
-		line-height: var(--typography-line-height-xl);
-	}
-}
+@import '../../styles/components/items-list';
```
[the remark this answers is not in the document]
damianmef: Moved.

lib/js/styles/ItemsList.vue
```diff
@@ -48,7 +49,7 @@ export default {
 	},
 	props: {
 		itemsLists: {
-			type: Array,
+			type: Array as PropType<Array<Array<Object>>>,
```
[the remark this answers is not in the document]
damianmef: Ok, I added the new type `ItemsListsItem` and used it here.

.github/workflows/ci.yml
```diff
@@ -24,6 +24,8 @@ jobs:
         run: yarn install --pure-lockfile
       - name: TypeScript check
         run: yarn ts:check
+      - name: Make envfile
```
kvas-damian, line 27: do we need this step?

damianmef: Now not, removed.

tools/importers/SynchronizeTypographyTokens.spec.ts
kvas-damian: you can perform assertions on input parameters in mocks: https://jestjs.io/docs/expect#tohavebeencalledwitharg1-arg2-
On the other hand I don't think it's that important

lib/js/styles/TokenTypes.ts
```diff
@@ -22,3 +24,9 @@ export interface TypographyToken {
 	attributes: Array<string>;
 	attributesRaw: Array<AttributeRaw>;
 }
+
+export interface ItemsListsItem {
+	title: string;
+	class: string;
+	list: Dict<object>;
```
kvas-damian, line 31: Why we can't specify type of this `object`?

kvas-damian (approved):

tools/importers/SynchronizeTypographyTokens.spec.ts
kvas-damian: I believe you can still perform the same tests without saving files on disk. Although, if it isn't that easy we can ignore it for now.

lib/js/styles/TokenTypes.ts
damianmef: I will think about this i other PR when I will fixing everything + improved views for UX.

PR #225, opened 2022-11-16
IT-4896 - Clean up

damianmef:


lib/styles/settings/_typography.scss
```diff
@@ -1,176 +1,199 @@
...
-$font-size-button-l: 14px;
-$line-height-button-l: 16px;
+$font-size-button-l-DEPRECATED: 14px;
+$line-height-button-l-DEPRECATED: 16px;
 
-@mixin textInfoS {
-	font-size: $font-size-info-s;
-	line-height: $line-height-info-s;
+@mixin textInfoS-DEPRECATED {
+	@warn "This mixin is deprecated!";
```
kvas-damian, line 62: Did you check how many worning will it generate on platform? Is it still posibble to read Vite output/stacktrace between these warnings?

damianmef: Right, it's redundant.

PR #230, opened 2022-11-25
IT-4933 tile ellipsis

pioziol123:


lib/js/components/Tile/Tile.vue
```diff
@@ -186,13 +187,25 @@ $tile-colors: (
...
+		}
 	}
 
 	&__text {
 		@include textM;
 
 		color: $color-neutral-text-heavy;
 		margin-top: $space-xxxxxs;
+
+		&.-ellipsis {
```
tweetgeek, line 204: we duplicate this code, maybe we should take it higher?

tweetgeek (approved):

PR #231, opened 2022-11-25
add radius to Dropdown

tweetgeek:


lib/js/components/Dropdown/Dropdown.consts.ts
```diff
@@ -2,3 +2,9 @@ export const DROPDOWN_TRIGGER_ACTIONS = {
 	CLICK: 'click',
 	HOVER: 'hover',
 } as const;
+
+export const DROPDOWN_RADIUS = {
```
rogatty, line 6: RADIUSES
https://github.com/bethinkpl/design-system/blob/dbfb47106a03925160a8c3038c153b4bb3e4404b/lib/js/components/Buttons/Button/Button.consts.ts#L21

rogatty (approved):

PR #235, opened 2022-12-01
IT-4895 - fixes

damianmef:
IT-4895 - IT-4939 - fix fontweight (split in variables view),
IT-4895 - IT-4941 - fix SectionHeader


lib/js/styles/ItemsTypography.vue
```diff
@@ -3,7 +3,7 @@
 		<div v-for="item in items" :key="item.id" class="itemsListItem">
 			<div v-if="item.label" class="itemsListItem__label">{{ item.label }}</div>
 			<div v-if="item.value" class="itemsListItem__value">{{ item.value }}</div>
-			<div v-if="item.ratio" class="itemsListItem__value">{{ item.ratio }}</div>
+			<div v-if="item.ratio || item.value" class="itemsListItem__value">{{ item.ratio }}</div>
```
kvas-damian, line 6: do we want to display empty div, when `ratio` is empty and `value` is set?

damianmef: Yes, because without this we have that case:
Screenshot 2022-12-01 at 14 17 36

lib/js/components/SectionHeader/SectionHeader.vue
```diff
@@ -46,6 +46,7 @@ $icons-and-slot-min-height-l: 50px;
 	display: flex;
 	flex-wrap: wrap;
 	justify-content: space-between;
+	margin: $space-xxxxxs 0;
```
kvas-damian, line 49: why do you use `margin` on top level element? We usually set margin on a component, to set correct space between items - in such scenario this margin will be overriten. It's safer to use padding.

lib/js/components/SectionHeader/SectionHeader.vue
```diff
@@ -84,23 +84,29 @@ $icons-and-slot-min-height-l: 50px;
...
 	}
 
+	&__slot,
+	&__iconWrapper {
+		display: flex;
+		align-items: center;
+	}
+
+	&__iconWrapper {
+		min-height: $icons-and-slot-min-height-m;
```
kvas-damian, line 94: why don't you center icon using flex?
`align-self: center` should do the trick

lib/js/components/SectionHeader/SectionHeader.vue
```diff
@@ -118,10 +124,15 @@ $icons-and-slot-min-height-l: 50px;
...
 		}
 
-		#{$root}__slot,
 		#{$root}__iconWrapper {
 			min-height: $icons-and-slot-min-height-l;
 		}
+
+		#{$root}__slot {
+			@media #{breakpoint-s()} {
+				min-height: $icons-and-slot-min-height-l;
```
kvas-damian, line 133: IMO, `height` should be set on `.sectionHeader` and rest should be centered using flex. Right now, you have to set height in multiple places

lib/styles/components/_items-list-item.scss
```diff
@@ -26,9 +26,15 @@
...
 
+	&__label {
+		flex: 1 1 auto;
+		width: 50%;
+	}
+
 	&__value {
 		color: var(--raw-gray-600);
 		flex: 1 1 auto;
+		width: 50%;
```
kvas-damian, line 37: it's funny that we use `50%` for 3-column layout. Did you consider using table for these pages?

lib/js/styles/ItemsTypography.vue
kvas-damian: https://github.com/bethinkpl/design-system/pull/235#discussion_r1037107320

Maybe `<table>` solves your problems?

lib/js/components/SectionHeader/SectionHeader.vue
damianmef: We can't, because of this case where text will have more than 1 line, it's UX recommendation:
Screenshot 2022-12-01 at 15 19 56

lib/js/components/SectionHeader/SectionHeader.vue
damianmef: Right, fixed.

lib/js/components/SectionHeader/SectionHeader.vue
damianmef: I think it's connected with above comment, where we need to center icons in the center when text has 1 line, but stick to the top when text has more than 1 line.

lib/styles/components/_items-list-item.scss
damianmef: I convert it to the table. I know it looks funny, but it works fine: if we have 2 columns or 3 they are always equal.

lib/js/styles/ItemsTypography.vue
damianmef: I converted.

lib/js/styles/ItemsTypography.vue
```diff
@@ -1,25 +1,25 @@
...
-			<div v-if="item.label" class="itemsListItem__label">{{ item.label }}</div>
-			<div v-if="item.value" class="itemsListItem__value">{{ item.value }}</div>
-			<div v-if="item.ratio" class="itemsListItem__value">{{ item.ratio }}</div>
-			<div v-if="item.token" class="itemsListItem__token">{{ item.token }}</div>
-			<div v-if="item.attributesRaw" class="itemsListItem__attributes">
+	<table class="itemsListItem">
+		<tr v-for="item in items" :key="item.id" class="itemsListItem__row">
+			<td v-if="item.label" class="itemsListItem__label">{{ item.label }}</td>
+			<td v-if="item.value" class="itemsListItem__value">{{ item.value }}</td>
+			<td v-if="item.ratio || item.value" class="itemsListItem__value">{{ item.ratio }}</td>
```
kvas-damian, line 6: do we still need to render empty `<td>`?

damianmef: I removed it just now.

PR #237, opened 2022-12-15
It 4909 popper slot build

pioziol123:


kvas-damian:
@pioziol123 can we close this PR, or is it something we'd like to merge. I see popper is already upgraded in `master`

PR #239, opened 2022-12-29
LMS-1525 | add subtitle props for popover

preziu:


lib/js/components/PopOver/PopOver.stories.ts
```diff
@@ -27,6 +27,7 @@ const args = {
 	color: POP_OVER_COLORS.DEFAULT,
 	triggerAction: POP_OVER_TRIGGER_ACTIONS.CLICK,
 	titleText: 'Lorem ipsum',
+	subtitleText: 'subLorem ipsum',
```
adamkarminski, line 30: `Dolor sit amet` ;)

kvas-damian (approved):

PR #251, opened 2023-02-13
IT-5036 | add disabled color to Pill

tweetgeek:


lib/js/components/Pill/Pill.vue
```diff
@@ -87,6 +132,19 @@ $pill-colors: (
 			#{$self}__label {
 				color: map-get($color-map, 'label');
 			}
+
+			&:disabled,
```
rogatty, line 136: I don't think this makes sense outside of form controls: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled

lib/js/components/Pill/Pill.vue
```diff
@@ -95,6 +153,12 @@ $pill-colors: (
 	display: inline-flex;
 	padding: $space-xxxxxs $space-xxxxxs $space-xxxxxs $space-xxs;
 
+	&:disabled,
+	&.-disabled {
+		cursor: not-allowed;
+		pointer-events: none;
```
rogatty, line 159: I think `cursor` is ignored when `pointer-events: none`

lib/js/components/Pill/Pill.vue
```diff
@@ -176,11 +244,16 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		isDisabled: {
```
rogatty, line 247: We've agreed to use `state` instead.

rogatty (approved):

PR #256, opened 2023-02-21
LMS-1525 | add props to PopOver.vue

preziu:


lib/js/components/PopOver/PopOver.vue
```diff
@@ -208,6 +209,16 @@ export default {
...
 		},
+		appendToBody: {
+			type: Boolean,
+			default: false,
+		},
+		modifiers: {
+			type: Object,
+			default() {
+				return {};
+			},
```
tweetgeek, line 220: You can use shortcut :) 
```suggestion
			default: () => ({}),
```

lib/js/components/PopOver/PopOver.vue
```diff
@@ -208,6 +209,16 @@ export default {
...
 			default: null,
 		},
+		appendToBody: {
+			type: Boolean,
+			default: false,
+		},
+		modifiers: {
+			type: Object,
+			default() {
+				return {};
```
rogatty, line 219: https://github.com/floating-ui/floating-ui/blob/v1.x/packages/popper/src/index.js#L40-L52
Empty object should be fine 👍

rogatty (approved):

PR #270, opened 2023-03-22
IT-5092 | Modal emits button click

kvas-damian:


lib/js/components/Modals/Modal/Modal.vue
```diff
@@ -88,6 +90,7 @@
 								class="ds-modal__buttonPrimary"
 								:color="calcFooterPrimaryButtonColor"
 								:icon-right="footerPrimaryButtonIcon"
+								@click="$emit('primary-button-click')"
```
rogatty, line 93: Don't we need `.native`?

kvas-damian: no, Button emits Vue event as well: https://github.com/bethinkpl/design-system/blob/d7fdfee5a2e6b806c08ece6263d45a0bb17362e9/lib/js/components/Buttons/Button/Button.vue#L24

PR #278, opened 2023-03-31
IT-5110 fix modal margins and secondary cta footer part

pioziol123:


lib/js/components/Modals/Modal/Modal.vue
```diff
@@ -264,53 +256,46 @@ $image-height-small: 140px;
...
 	}
 
-	&__buttonPrimary {
-		&:not(:first-child) {
-			// first-child because of reverse order in flex-direction
-			margin-bottom: $space-s;
+	&__footerCtaSecondary {
+		align-items: center;
+		display: flex;
+		flex-direction: row;
```
kvas-damian, line 293: it's a default value for flex. Do we need it?

lib/js/components/Modals/Modal/Modal.vue
```diff
@@ -264,53 +256,46 @@ $image-height-small: 140px;
...
+		justify-content: space-between;
 
-			@media #{breakpoint-s()} {
-				margin-bottom: 0;
-			}
+		@media #{breakpoint-s()} {
+			gap: 0 $space-m;
```
kvas-damian, line 297: don't we need to define this gap also for mobile, to reserve minimal spacing between checkbox and button when labels are long?

lib/js/components/Modals/Modal/Modal.vue
kvas-damian, line 293:
```suggestion
```

lib/js/components/Modals/Modal/Modal.vue
kvas-damian: [`c510547` (#278)](https://github.com/bethinkpl/design-system/pull/278/commits/c51054708fce3d4d40a8cd29120373183531e49a)

kvas-damian (approved):

PR #285, opened 2023-04-19
add labelTextEllipsis for ProgressBar

tweetgeek:


lib/js/components/ProgressBar/ProgressBar.vue
```diff
@@ -8,7 +8,11 @@
 		<div v-if="labelText || labelDataExists" class="progressBar__label">
 			<div
 				class="progressBar__labelText"
-				:class="{ '-medium': labelTextSize === PROGRESS_BAR_LABEL_TEXT_SIZES.MEDIUM }"
+				:class="{
+					'-medium': labelTextSize === PROGRESS_BAR_LABEL_TEXT_SIZES.MEDIUM,
+					'-ellipsis': labelTextEllipsis,
+				}"
+				:title="labelTextEllipsis ? labelText : ''"
```
kvas-damian, line 15: if you pass `null` or `undefined` (don't remember which) then Vue won't apply this attribute to HTML

tweetgeek: i checked and it can be null or undefined

PR #291, opened 2023-04-28
IT-5156 thick scrollbar under s breakpoint

pioziol123:


lib/styles/design-system.scss
```diff
@@ -50,7 +50,9 @@ textarea {
 }
 
 html.thick-scrollbar {
-	@include scrollbar(15px, var(--raw-gray-100));
+	@media #{breakpoint-s()} {
```
tweetgeek, line 53: what about the tablet?

pioziol123: right :)
https://bethinkteam.slack.com/archives/C012R8RM3A4/p1682688253072869

tweetgeek: > Dopytać UX co z tabletem

PR #294, opened 2023-05-08
SectionHeader

tweetgeek:


lib/js/components/Headers/SectionHeader/SectionHeader.vue
```diff
@@ -0,0 +1,343 @@
...
+		padding: $space-xxxs $space-xxs;
+	}
+
+	&.-size-l {
+		#{$root}__titleWrapper {
+			@include heading-l-default-bold;
+		}
+
+		#{$root}__iconLeft {
+			margin-right: $space-xxs;
```
pioziol123, line 147: cannot we set column-gap  property for parent instead of margins in children?

lib/js/components/Headers/SectionHeader/SectionHeader.vue
```diff
@@ -0,0 +1,343 @@
...
+			default: false,
+		},
+		size: {
+			type: String,
+			default: SECTION_HEADER_SIZES.M,
+			validator(size) {
+				return Object.values(SECTION_HEADER_SIZES).includes(size);
+			},
+		},
+		title: {
```
pioziol123, line 286: I wonder if we should validate if length of the string is greater than 0.

lib/js/components/Headers/SectionHeader/SectionHeader.vue
```diff
@@ -0,0 +1,343 @@
...
+		:class="{
+			'-expandable': expandable,
+			'-horizontal': mobileLayout === SECTION_HEADER_MOBILE_LAYOUTS.HORIZONTAL,
+			[sizeClass]: true,
+		}"
+	>
+		<div class="sectionHeader__wrapper">
+			<div class="sectionHeader__main">
+				<div class="sectionHeader__titleWrapper" @click="onTitleWrapperClicked">
+					<div v-if="iconLeft" class="sectionHeader__iconWrapper sectionHeader__iconLeft">
```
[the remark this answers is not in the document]
pioziol123: is this div necessary? cannot we add `v-if` directive and `sectionHeader__iconLeft` to ds-icon component?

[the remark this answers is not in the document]
pioziol123: same with __iconRight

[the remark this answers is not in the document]
pioziol123: and expandable icon

lib/js/components/Headers/SectionHeader/SectionHeader.vue
pioziol123: if you remove margins from `__iconLeft` and `__iconRight` both selectors can be reduced to common `__icon` selector

lib/js/components/Headers/SectionHeader/SectionHeader.vue
```diff
@@ -0,0 +1,343 @@
...
+				</div>
+			</div>
+
+			<div v-if="supportingText" class="sectionHeader__supportingText"
+				>{{ supportingText }}
+			</div>
+
+			<div
+				v-if="$slots.default && showSlot"
+				class="sectionHeader__slotVertical -showOnMobile"
```
pioziol123, line 57: No such class as `-showOnMobile`

lib/js/components/Headers/SectionHeader/SectionHeader.vue
```diff
@@ -0,0 +1,343 @@
...
+			ICONS: Object.freeze(ICONS),
+			ICON_SIZES: Object.freeze(ICON_SIZES),
+			ICON_BUTTON_SIZES: Object.freeze(ICON_BUTTON_SIZES),
+			ICON_BUTTON_COLORS: Object.freeze(ICON_BUTTON_COLORS),
+			SECTION_HEADER_MOBILE_LAYOUTS: Object.freeze(SECTION_HEADER_MOBILE_LAYOUTS),
+		};
+	},
+	computed: {
+		chevronRotation(): number | null {
+			return this.isExpanded ? 180 : null;
```
pioziol123, line 315:
```suggestion
			return this.isExpanded ? 180 : 0;
```

lib/js/components/Headers/SectionHeader/SectionHeader.vue
pioziol123: same with `-hideOnMobile`

lib/js/components/Headers/SectionHeader/SectionHeader.vue
```diff
@@ -0,0 +1,343 @@
+<template>
+	<div
+		class="sectionHeader"
+		:class="{
+			'-expandable': expandable,
+			'-horizontal': mobileLayout === SECTION_HEADER_MOBILE_LAYOUTS.HORIZONTAL,
+			[sizeClass]: true,
+		}"
+	>
+		<div class="sectionHeader__wrapper">
```
pioziol123, line 10: can we merge `secrionHeader` and `__wrapper`. I see there is a divider but `__wrapper` has `flex-direction: column` anyway.

lib/js/components/Headers/SectionHeader/SectionHeader.vue
```diff
@@ -0,0 +1,343 @@
...
+		align-items: center;
+		display: flex;
+		width: 100%;
+
+		@media #{breakpoint-s()} {
+			width: auto;
+		}
+	}
+
+	&.-expandable &__titleWrapper {
```
pioziol123, line 110: Does it make sense to merge this selector with succeeding like
```
&.-expandable &__titleWrapper {
...
&:hover {
...
}
}
```

lib/js/components/Headers/SectionHeader/SectionHeader.vue
```diff
@@ -0,0 +1,343 @@
...
+		justify-content: space-between;
+		padding: $space-xxxxs 0;
+	}
+
+	&__supportingText {
+		@include text-s-compact-regular;
+
+		color: $color-neutral-text;
+		gap: $space-xxs;
+		padding: 0 0 $space-xs;
```
pioziol123, line 97: just padding-bottom?

lib/js/components/Headers/SectionHeader/SectionHeader.vue
```diff
@@ -0,0 +1,343 @@
...
+
+	&__iconLeft,
+	&__iconRight,
+	&__chevron {
+		color: $color-neutral-icon;
+	}
+
+	&__title {
+		color: $color-neutral-text-strong;
+		padding: $space-xxs 0;
```
pioziol123, line 134: can this padding be added to `__main` selector's padding?

pioziol123: Ah. I see it is to keep space between info as well. So maybe we should use `row-gap` in `__main` as well

pioziol123: or add padding from `__main` to this one

lib/js/components/Headers/SectionHeader/SectionHeader.vue
```diff
@@ -0,0 +1,343 @@
...
+		padding: 0;
+	}
+
+	&__main {
+		align-items: center;
+		display: flex;
+		flex-direction: row;
+		gap: $space-xxs;
+		justify-content: space-between;
+		padding: $space-xxxxs 0;
```
pioziol123, line 89: this padding changes with size. same for `__title` selector

lib/js/components/Headers/PageHeader/PageHeader.vue
```diff
@@ -0,0 +1,128 @@
...
+		mobileLayout: {
+			type: String,
+			default: PAGE_HEADER_MOBILE_LAYOUTS.VERTICAL,
+			validator: (value) => Object.values(PAGE_HEADER_MOBILE_LAYOUTS).includes(value),
+		},
+	},
+	data() {
+		return {
+			ICONS: Object.freeze(ICONS),
+			ICON_SIZES: Object.freeze(ICON_SIZES),
```
pioziol123, line 123: unused

lib/js/components/Headers/PageHeader/PageHeader.vue
```diff
@@ -0,0 +1,128 @@
...
+		},
+		mobileLayout: {
+			type: String,
+			default: PAGE_HEADER_MOBILE_LAYOUTS.VERTICAL,
+			validator: (value) => Object.values(PAGE_HEADER_MOBILE_LAYOUTS).includes(value),
+		},
+	},
+	data() {
+		return {
+			ICONS: Object.freeze(ICONS),
```
pioziol123, line 122: unused

lib/js/components/Headers/SectionHeader/SectionHeader.vue
tweetgeek: Why? we have `required: true`, if someone pass an empty string then that's his problem

lib/js/components/Headers/SectionHeader/SectionHeader.vue
tweetgeek: if we use `null`, vue don't render this prop and use default value if we pass 0 and in future we change default value it blowup :D

pioziol123: IDE is underlining the line where this property is used. Another thing is why we may want to change meaning of 0 for this prop, it would be counterintuitive.

lib/js/components/Headers/SectionHeader/SectionHeader.vue
tweetgeek: yes

lib/js/components/Headers/SectionHeader/SectionHeader.vue
tweetgeek: if we pass `0` we got an error from the validator because we have `return [90, 180, 270].includes(value)`

tweetgeek: no longer underlined, changed to `undefined`

lib/js/components/Headers/SectionHeader/SectionHeader.vue
pioziol123: I guess :)

pioziol123 (approved):

lib/js/components/Headers/SectionHeader/SectionHeader.stories.ts
```diff
@@ -44,32 +49,60 @@ const StoryTemplate: StoryFn<typeof SectionHeader> = (args, { updateArgs }) => (
...
+	size: SECTION_HEADER_SIZES.MEDIUM,
+	title: 'Section Header',
+	supportingText:
+		'Supporting text. Et doloribus aspernatur suscipit provident maiores. Natus natus et pariatur. Eligendi illo quo esse. Tenetur ad neque veniam.',
+	eyebrow: 'eyebrow text',
+	iconLeft: ICONS.FA_PLAY,
+	iconRight: ICONS.FA_BOOK,
+	info: true,
+	divider: true,
 	expandable: false,
```
pioziol123, line 61: isExpandable

lib/js/components/Headers/SectionHeader/SectionHeader.vue
```diff
@@ -0,0 +1,383 @@
...
+	}
+
+	&__supportingText {
+		@include text-s-compact-regular;
+
+		color: $color-neutral-text;
+		padding: 0 0 $space-xs;
+
+		@media #{breakpoint-s()} {
```
pioziol123, line 103: why not on mobile? i think this padding should be removed on mobile as well when the mobileLayout is horizontal

lib/js/components/Headers/SectionHeader/SectionHeader.vue
```diff
@@ -156,6 +169,10 @@
 		#{$root}__titleWrapper {
 			gap: $space-xxs;
 		}
+
+		#{$root}__info {
+			padding: $space-xxs 0 $space-xxs $space-xxs;
```
pioziol123, line 192: why left padding? ewa mentioned only top/bottom

lib/js/components/Headers/SectionHeader/SectionHeader.vue
tweetgeek: Because on mobile we display slot below this text, so we need padding

tweetgeek: fuck..

PR #295, opened 2023-05-11
DSLayout

tweetgeek:
https://wnl-design-system-preview.s3.amazonaws.com/IT-5149/index.html?path=/story/components-layout--interactive


lib/js/components/Layout/Layout.vue
```diff
@@ -0,0 +1,261 @@
...
+	LAYOUT_CONTENT_COLUMN_SIZE,
+	LAYOUT_RIGHT_COLUMN_MODE,
+	LAYOUT_RIGHT_COLUMN_SIZE,
+	LayoutContentColumnSize,
+	LayoutRightColumnMode,
+	LayoutRightColumnSize,
+} from './Layout.consts';
+
+export default {
+	name: 'Layout',
```
kvas-damian, line 206: This component name looks too generic. What if we want to introduce another type of layout in future?

lib/js/components/Layout/Layout.vue
```diff
@@ -0,0 +1,261 @@
+<template>
+	<div class="layout">
+		<div v-if="sideMenuVisible" class="layout__navigation">
+			<div class="layout__mainMenu">
+				<slot name="mainMenu" />
+			</div>
+			<div class="layout__sideNav">
+				<slot name="sideNav" />
+			</div>
```
kvas-damian, line 9: did UX agree on separating left column into `mainMenu` and `sideNav`? How do we plan to implement `AssessmentPlanEditor` when `mainMenu` has fixed width?

lib/js/components/Layout/Layout.consts.ts
```diff
@@ -0,0 +1,22 @@
+export const LAYOUT_CONTENT_COLUMN_SIZE = {
+	FULL_WIDTH: 'fullWidth',
+	MAX_900: 'max900',
```
kvas-damian, line 3: using pixel value in const name requires migration of all component usages when we decide to change size. IMO, we should avoid that

lib/js/components/Layout/Layout.vue
```diff
@@ -0,0 +1,261 @@
...
+		</div>
+		<div
+			class="layout__contentColumn"
+			:class="{
+				'-fullWidth': isContentFullWidth,
+				'-max900': !isContentFullWidth,
```
kvas-damian, line 15: similar to https://github.com/bethinkpl/design-system/pull/295/files#r1193359026
don't use pixel value in class name

lib/js/components/Layout/Layout.vue
```diff
@@ -0,0 +1,261 @@
...
+			@media #{breakpoint-l()} {
+				max-width: $right-column-medium-l-max-width;
+				min-width: $right-column-medium-l-min-width;
+				width: $right-column-medium-l-width;
+			}
+		}
+
+		&.-large {
+			max-width: $right-column-large-max-width;
+			width: 100%;
```
kvas-damian, line 177: do you need to override value defined in `&__rightColumn`?

lib/js/components/Layout/Layout.vue
```diff
@@ -0,0 +1,261 @@
...
+
+		@media #{breakpoint-l()} {
+			height: auto;
+			position: initial;
+			right: initial;
+			top: initial;
+		}
+
+		&.-medium {
+			max-width: $right-column-medium-max-width;
```
kvas-damian, line 165: according to my knowledge right column should be 100% on mobile (<760px). Right now overlay is not displayed so on 600px width resolution you see broken content

lib/js/components/Layout/Layout.vue
```diff
@@ -0,0 +1,261 @@
...
+			default: LAYOUT_RIGHT_COLUMN_MODE.COLUMN_VISIBLE,
+			validator(rightColumnMode) {
+				return Object.values(LAYOUT_RIGHT_COLUMN_MODE).includes(rightColumnMode);
+			},
+		},
+		sideBarVisible: {
+			type: Boolean,
+			default: false,
+		},
+		rightColumnVisible: {
```
kvas-damian, line 233: With a single prop it could be hard to determine default visibility state on mobile. You cannot use CSS media queries on mobile, but you have to inform component if column should be open (desktop) or closed (mobile) on initial render.

@rogatty do you remember if we already found a solution for that?

lib/js/components/Layout/Layout.stories.ts
```diff
@@ -0,0 +1,94 @@
...
+} as Meta<typeof Layout>;
+
+const StoryTemplate: StoryFn<typeof Layout> = (args) => ({
+	components: { Layout },
+	setup() {
+		return { ...args };
+	},
+	template: `
+		<div
+			style='height: 100%; width: 100%; display: flex; min-height: 100vh; border: 1px solid gray; font-size: 12px; color: #fff;'>
```
kvas-damian, line 22: you can add use `layout: 'fullscreen'` to avoid default stroybook paddings in this story

lib/js/components/Layout/Layout.vue
```diff
@@ -0,0 +1,261 @@
...
+			display: block;
+		}
+
+		@media #{breakpoint-l()} {
+			display: none;
+		}
+	}
+
+	&__navigation {
+		display: none;
```
kvas-damian, line 108: it isn't possible to show `__navigation` now on mobile. If I remember correctly, we agreed to show it on an overlay

lib/js/components/Layout/Layout.vue
```diff
@@ -0,0 +1,261 @@
...
+					'-large': rightColumnSize === LAYOUT_RIGHT_COLUMN_SIZE.LARGE,
+				}"
+			>
+				<slot name="rightColumn" />
+			</div>
+		</template>
+		<div
+			v-if="
+				$slots.sideBar &&
+				sideBarVisible &&
```
kvas-damian, line 43: I'm not sure if we agreed with UX that collapsed right collumn will be rendered using separate slot or different way. Do you have a confirmation for that?

lib/js/components/Layout/Layout.vue
```diff
@@ -0,0 +1,261 @@
...
+
+$side-bar-width: 66px;
+
+.layout {
+	$root: &;
+
+	display: flex;
+	flex-wrap: nowrap;
+	height: auto;
+	justify-content: space-between;
```
kvas-damian, line 86: IMO, there should never be spacing between columns. Why do we need this?

lib/js/components/Layout/Layout.vue
```diff
@@ -0,0 +1,261 @@
...
+	padding: 0;
```
kvas-damian, line 87: do we need to override default browser value?

lib/js/components/Layout/Layout.vue
```diff
@@ -0,0 +1,261 @@
...
+	position: relative;
+	width: 100%;
+
+	&__overlay {
+		background: rgba(0, 0, 0, 0.7);
```
kvas-damian, line 92: color token?

lib/js/components/Layout/Layout.vue
```diff
@@ -0,0 +1,261 @@
...
+
+		@media #{breakpoint-l()} {
+			display: none;
+		}
+	}
+
+	&__navigation {
+		display: none;
+		flex-wrap: nowrap;
+		justify-content: flex-start;
```
kvas-damian, line 110: does it change anything?

lib/js/components/Layout/Layout.vue
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	&__mainMenu {
+		height: auto;
+		width: $main-menu-width;
+	}
+
+	&__sideNav {
+		height: auto;
+		width: 100%;
```
kvas-damian, line 127: 100%? We already show `__mainMenu` in this container, so probably there is not enough space to render both

lib/js/components/Layout/Layout.vue
```diff
@@ -0,0 +1,261 @@
...
+		margin: 0 auto;
+		padding: $space-s;
+		width: auto;
+
+		&.-noPadding {
+			padding: 0;
+		}
+
+		&.-max900 {
+			max-width: $content-column-max-width;
```
kvas-damian, line 141: this column should always take all available space to allow scrolling when cursor is over white margins. Only content inside the column should have limited width

lib/js/components/Layout/Layout.consts.ts
```diff
@@ -0,0 +1,22 @@
...
+} as const;
+
+export type LayoutContentColumnSize =
+	typeof LAYOUT_CONTENT_COLUMN_SIZE[keyof typeof LAYOUT_CONTENT_COLUMN_SIZE];
+export const LAYOUT_RIGHT_COLUMN_SIZE = {
+	MEDIUM: 'medium',
+	LARGE: 'large',
```
kvas-damian, line 10: IMO, `LARGE` looks like a separate layout, can you discuss with UX if we should expose this prop for 3 column layout?

lib/js/components/Layout/Layout.vue
rogatty: I don't remember talking about that. I think we can assume columns should always be closed on mobile on initial render. If that's true, then I think we can make it work without adding a new prop, local data field should be enough. Mobile would take into consideration only the local one while desktop would use both, something like `:class={'-desktopVisible': sideMenuVisible || sideMenuVisibleLocal, '-mobileVisible': sideMenuVisibleLocal}`

lib/js/components/Layout/Layout.vue
```diff
@@ -0,0 +1,261 @@
...
+			},
+		},
+		rightColumnSize: {
+			type: String as PropType<LayoutRightColumnSize>,
+			default: LAYOUT_RIGHT_COLUMN_SIZE.MEDIUM,
+			validator(rightColumnSize) {
+				return Object.values(LAYOUT_RIGHT_COLUMN_SIZE).includes(rightColumnSize);
+			},
+		},
+		rightColumnMode: {
```
kvas-damian, line 222: there are two props `rightColumnMode and sideBarVisible` that I have to change to render `sideBar`, but I still can't get it rendered in storybook. Am I doing something wrong?

lib/js/components/Layouts/ThreeColumnLayout/ThreeColumnLayout.vue
```diff
@@ -0,0 +1,237 @@
...
+				display: block;
+			}
+		}
+
+		@media #{breakpoint-l()} {
+			display: none;
+		}
+	}
+
+	&__navigation {
```
kvas-damian, line 89: can we unify class name and slot name?

kvas-damian: and prop name

lib/js/components/Layouts/ThreeColumnLayout/ThreeColumnLayout.vue
```diff
@@ -0,0 +1,239 @@
+<template>
+	<div class="layout">
```
kvas-damian, line 2: should we change class name to match component name?

lib/js/components/Layouts/ThreeColumnLayout/ThreeColumnLayout.vue
```diff
@@ -0,0 +1,239 @@
...
+			position: initial;
+			top: initial;
+		}
+
+		&.-mobileVisible {
+			display: initial;
+		}
+	}
+
+	&__contentColumn {
```
kvas-damian, line 124: there is no way to scroll content in this container. I'm wondering if we need to fix it also for other columns or not?

lib/js/components/Layout/Layout.vue
kvas-damian: there is something to be changed here. Definietely prop names are confusing

kvas-damian: I'm still not sure if it'll cover all our use-cases, but let's check that during the implementation

lib/styles/settings/_z-indexes.scss
```diff
@@ -1,3 +1,4 @@
 // z-index positioning
 // hardcoded
 $z-index-modal: 16777270; // 16777271 is the highest possible value in Safari, we need to make space for elements that above modal, so -1
+$z-index-overlay: 51; // 51 based on settings from platform $z-index-overlay: $z-index-top-nav + 1;, where $z-index-top-nav = 50
```
kvas-damian, line 4: it should be other way round. `wnl-platform` depends on DS, so we should add this comment there

lib/js/components/Layouts/ThreeColumnLayout/ThreeColumnLayout.vue
```diff
@@ -0,0 +1,265 @@
...
+
+	&__overlay {
+		background: $color-default-overlay;
+		display: none;
+		height: 100%;
+		left: 0;
+		position: fixed;
+		top: 0;
+		width: 100%;
+		z-index: $z-index-overlay + 1;
```
kvas-damian, line 84: Why do we need to increase default `z-index` value? Should we add 1 in the variable?

kvas-damian (approved):

PR #306, opened 2023-05-29
IT-5075 | RadioButton and Checkbox

tweetgeek:


lib/js/components/Form/Toggle/Toggle.vue
```diff
@@ -0,0 +1,305 @@
...
+			'icon': $color-neutral-icon-disabled,
+		),
+		'selected': (
+			'label': $color-neutral-text-heavy-disabled,
+			'icon': $color-primary-icon-disabled,
+		),
+	),
+	'loading': (
+		'not-selected': (
+			'label': $color-neutral-text-heavy-disabled,
```
kvas-damian, line 67: loading should like `default`, not `disabled`: https://miro.com/app/board/o9J_knPnexA=/?moveToWidget=3458764546392448081&cot=14

lib/js/components/Form/SelectionControl/SelectionControl.vue
```diff
@@ -0,0 +1,352 @@
...
+	&__labelWrapper {
+		display: flex;
+	}
+
+	&__icon {
+		position: relative;
+	}
+
+	&.-x-small {
+		#{$root}__label {
```
kvas-damian, line 229: label isn't correctly positioned for xSmall size

lib/js/components/Form/SelectionControl/SelectionControl.vue
```diff
@@ -0,0 +1,352 @@
+<template>
+	<div
```
kvas-damian, line 2: most likely, if you convert this `div` into a `label` then you could remove `click` handler. Clicks on a `label` tag automagically toggles checkbox/radio

lib/js/components/Form/SelectionControl/SelectionControl.vue
```diff
@@ -0,0 +1,352 @@
...
+	@each $name, $map in $selection-control-colors {
+		@if map-get($map, 'hoverable') {
+			&.-#{$name} {
+				@each $size, $map in $selection-control-sizes {
+					&.-#{$size} {
+						#{$root}__icon::before {
+							bottom: -#{map-get($map, 'iconBorder')};
+							left: -#{map-get($map, 'iconBorder')};
+							right: -#{map-get($map, 'iconBorder')};
+							top: -#{map-get($map, 'iconBorder')};
```
kvas-damian, line 106: it looks more like an outline size. Maybe `iconOutlinePadding`?

lib/js/components/Form/SelectionControl/SelectionControl.vue
```diff
@@ -0,0 +1,352 @@
...
+					background-color: transparent;
+					border-radius: 50%;
+					content: '';
+					position: absolute;
+					transform: scale(0);
+					transition: all $default-transition-time ease 0ms;
+					z-index: -1;
+				}
+
+				&:hover,
```
kvas-damian, line 123: I'm not convinced that defining these styles in for loop makes it easier to understand. What would you say about moving it into regular styles block

lib/js/components/Form/SelectionControl/SelectionControl.vue
```diff
@@ -0,0 +1,352 @@
...
+		@click="onToggle"
+	>
+		<div class="selectionControl__iconWrapper">
+			<icon :icon="icon" :size="iconSize" class="selectionControl__icon" />
+		</div>
+		<div v-if="label" class="selectionControl__labelWrapper">
+			<div class="selectionControl__label">{{ label }}</div>
+		</div>
+		<input
+			type="checkbox"
```
kvas-damian, line 23: should it be `checkbox` when we use it to render radio?

lib/js/components/Form/SelectionControl/SelectionControl.vue
tweetgeek: but it is not good practice to use div in label

kvas-damian: why don't you use `spans` inside label?

lib/js/components/Form/SelectionControl/SelectionControl.vue
kvas-damian: not checkbox/radio is centered when label goes into multiline

kvas-damian:
> I'm wondering if we should implement something similar to `v-radio-group` https://vuetifyjs.com/en/components/radio-buttons/ to make sure that logic which deselects other selected radios is in one place and tested well.
> 
> Also we could think about improving component API by adopting features from vuetify: https://vuetifyjs.com/en/components/checkboxes/#model-as-array and https://vuetifyjs.com/en/components/checkboxes/#model-as-boolean

we agreed offline, that we can do that we can implement it in future

lib/js/components/Form/SelectionControl/SelectionControl.vue
```diff
@@ -0,0 +1,339 @@
...
+			<icon :icon="icon" :size="iconSize" class="selectionControl__icon" />
+		</div>
+		<div v-if="label" class="selectionControl__labelWrapper">
+			<div class="selectionControl__label">{{ label }}</div>
+		</div>
+		<input
+			:type="type === SELECTION_CONTROL_TYPE.CHECKBOX ? 'checkbox' : 'radio'"
+			class="selectionControl__checkbox"
+			:value="isSelected"
+			@input="onToggle"
```
kvas-damian, line 26: why did we change it?

kvas-damian (approved):

PR #307, opened 2023-06-02
IT-5074 | SelectionTile

tweetgeek:


lib/js/components/SelectionTile/SelectionTile.vue
```diff
@@ -0,0 +1,258 @@
...
+
+		color: $color-neutral-text;
+	}
+
+	&__icon {
+		color: $color-neutral-icon;
+		display: flex;
+	}
+
+	&:hover:not(.-loading) {
```
kvas-damian, line 113: should we trigger `hover/focused` state when navigating to Radio using a keybord?

lib/js/components/SelectionTile/SelectionTile.vue
```diff
@@ -0,0 +1,258 @@
...
+
+	align-items: center;
+	background-color: $color-neutral-background-weak;
+	border-radius: $radius-s;
+	column-gap: $space-xxs;
+	cursor: pointer;
+	display: flex;
+	justify-content: space-between;
+	min-height: $selection-tile-min-height;
+	outline: 1px solid $color-neutral-border;
```
kvas-damian, line 72: why do we have to use `outline` instead of a `border`?

lib/js/components/SelectionTile/SelectionTile.vue
```diff
@@ -0,0 +1,258 @@
...
+				'-selected': isSelected,
+				'-default': state === SELECTION_TILE_STATE.DEFAULT,
+				'-disabled': state === SELECTION_TILE_STATE.DISABLED,
+				'-loading': state === SELECTION_TILE_STATE.LOADING,
+			},
+		]"
+		@click="updateIsSelected(!isSelected)"
+	>
+		<div class="selectionTile__wrapper">
+			<div class="selectionTile__selectionControl">
```
kvas-damian, line 15: Do we need this wrapper? We could set this class on `radio/checkbox`

lib/js/components/SelectionTile/SelectionTile.vue
```diff
@@ -0,0 +1,258 @@
...
+
+				<div v-if="supportingText" class="selectionTile__supportingText"
+					>{{ supportingText }}
+				</div>
+			</div>
+		</div>
+		<div
+			v-if="icon || isLoading"
+			class="selectionTile__icon"
+			@click.stop="isLoading ? null : $emit('icon-click')"
```
kvas-damian, line 42: why don't we move these attributes directly to `icon` tag?

lib/js/components/SelectionTile/SelectionTile.vue
```diff
@@ -0,0 +1,258 @@
...
+	align-items: center;
+	background-color: $color-neutral-background-weak;
+	border-radius: $radius-s;
+	column-gap: $space-xxs;
+	cursor: pointer;
+	display: flex;
+	justify-content: space-between;
+	min-height: $selection-tile-min-height;
+	outline: 1px solid $color-neutral-border;
+	padding: $space-xs;
```
kvas-damian, line 73: there is something wrong with right icon position because of this padding, see discussion in Figma: https://www.figma.com/file/izQdYyiBR1GQgFkaOIfIJI?node-id=6085:110922#462383300

lib/js/components/SelectionTile/SelectionTile.vue
```diff
@@ -0,0 +1,258 @@
...
+		color: $color-neutral-text;
+	}
+
+	&__icon {
+		color: $color-neutral-icon;
+		display: flex;
+	}
+
+	&:hover:not(.-loading) {
+		background-color: $color-neutral-background-hovered;
```
kvas-damian, line 114: weak-hovered?

lib/js/components/SelectionTile/SelectionTile.vue
```diff
@@ -0,0 +1,258 @@
...
+			ICONS: Object.freeze(ICONS),
+			ICON_SIZES: Object.freeze(ICON_SIZES),
+			RADIO_BUTTON_SIZE: Object.freeze(RADIO_BUTTON_SIZE),
+			SELECTION_TILE_STATE: Object.freeze(SELECTION_TILE_STATE),
+			SELECTION_TILE_TYPE: Object.freeze(SELECTION_TILE_TYPE),
+			CHECKBOX_SIZE: Object.freeze(CHECKBOX_SIZE),
+		};
+	},
+	computed: {
+		selectionControlState() {
```
kvas-damian, line 239: instead of this computed, maybe we should have a universal state for both radio and checkbox?

lib/js/components/SelectionTile/SelectionTile.vue
```diff
@@ -0,0 +1,258 @@
...
+		</div>
+		<div
+			v-if="icon || isLoading"
+			class="selectionTile__icon"
+			@click.stop="isLoading ? null : $emit('icon-click')"
+		>
+			<icon
+				:icon="isLoading ? ICONS.FAD_SPINNER_THIRD : icon"
+				:size="ICON_SIZES.X_SMALL"
+				:spinning="state === SELECTION_TILE_STATE.LOADING"
```
kvas-damian, line 47:
```suggestion
				:spinning="isLoading"
```
to make it consisten with consition in `:icon`

lib/js/components/SelectionTile/SelectionTile.vue
kvas-damian, line 42: Minor issue, I found this because I was experimenting with storybok Actions Tab
because of `.stop` here, `click` event isn't fired on entire component. I'd expect `click` to be fired here as well...

lib/js/components/SelectionTile/SelectionTile.vue
tweetgeek: Because the border changes the height of this component and the outline not, so in the base version with only title we always got 48px not 50.

lib/js/components/SelectionTile/SelectionTile.vue
tweetgeek: stop here is for stopping bubbling when clicking this icon, without this after click in icon (eg to open filters) we're toggling selection tile, i changed it to no stopping when loading

lib/js/components/SelectionTile/SelectionTile.vue
tweetgeek: but we showing it

lib/js/components/SelectionTile/SelectionTile.vue
kvas-damian: Figma is fixed now

lib/js/components/SelectionTile/SelectionTile.vue
kvas-damian: Zaznaczenie_146
I see focus only on radio, not entire SelectionTile

lib/js/components/SelectionTile/SelectionTile.vue
kvas-damian: I was also referring to class name and `if` :)

lib/js/components/SelectionTile/SelectionTile.vue
```diff
@@ -0,0 +1,250 @@
...
+	>
+		<div class="selectionTile__wrapper">
+			<radio-button
+				v-if="type === SELECTION_TILE_TYPE.RADIO_BUTTON"
+				:size="RADIO_BUTTON_SIZE.X_SMALL"
+				:is-selected="isSelected"
+				:state="SELECTION_CONTROL_STATE_MAP[state]"
+				@update:isSelected="updateIsSelected($event)"
+			/>
+			<checkbox
```
kvas-damian, line 22: Radio and Checkbox components has the same API right now. We could use `<component is="">` to guarantee noone forgets to update 2nd component when introducing changes in the first one

lib/js/components/SelectionTile/SelectionTile.vue
kvas-damian: ok, let's keep it as it is. I hope noone uses `click` event handler on this component :see_no_evil:

kvas-damian (approved):

PR #310, opened 2023-07-19
Fix DrawerTile spacing

kvas-damian:


lib/js/components/Drawer/DrawerTile/DrawerTile.vue
```diff
@@ -1,5 +1,7 @@
 <template>
-	<ds-tile v-bind="$attrs" class="drawerTile" />
+	<div class="drawerTile">
```
kvas-damian, line 2: now `Tile` props are injected as HTML attributes on this div :(

lib/js/components/Drawer/DrawerTile/DrawerTile.vue
```diff
@@ -1,5 +1,7 @@
...
+		<ds-tile v-bind="$attrs" />
+	</div>
```
rogatty, line 4: It emits `click` event even if disabled. It's the same on master but maybe it's a good opportunity to fix it?

This is ignored:
https://github.com/bethinkpl/design-system/blob/33ebab65073bc127bb8b0fc516e4c42ce864fbe9/lib/js/components/Tile/Tile.vue#L171

rogatty (approved):

lib/js/components/Drawer/DrawerTile/DrawerTile.vue
rogatty: It's not the first place where we do this. I'm not sure if we should care. It's caused by https://vuejs.org/guide/components/attrs.html#attribute-inheritance

lib/js/components/Drawer/DrawerTile/DrawerTile.vue
kvas-damian: https://github.com/bethinkpl/design-system/pull/310/commits/f68b23b668af14803d0be0448a4bdc62f4dae94a

lib/js/components/Drawer/DrawerTile/DrawerTile.vue
kvas-damian: there are more problems with not defined props on this component... 
Because of `v-bind="$attrs"` when you use `<drawer-tile class="test">` you get this class name applied on both DrawerTile and Tile containers...

kvas-damian: https://github.com/bethinkpl/design-system/pull/310/commits/7eaf5c38e3d5427e5dcb8807d5a4d714a60ee18e

PR #329, opened 2023-12-20
IT-5525 | Allow no elaboration in SurveyQuestionScale

rogatty:


lib/js/components/SurveyQuestions/SurveyQuestionScale/SurveyQuestionScale.vue
```diff
@@ -205,7 +205,7 @@ export default {
 		},
 		elaborationLabel: {
 			type: String,
-			required: true,
+			default: null,
```
kvas-damian, line 208: should we update storybook to reflect default state of component?

rogatty: I don't think so, it's already far from the default.

kvas-damian (approved):

PR #336, opened 2024-01-18
add RichListGroupItem

tweetgeek:


lib/js/components/RichList/RichListGroupItem/RichListGroupItem.vue
```diff
@@ -0,0 +1,122 @@
...
+			},
+		},
+		borderColorHex: {
+			type: String,
+			default: null,
+		},
+	},
+	computed: {
+		borderColorClass() {
+			if (!this.borderColor || this.borderColorHex) {
```
pioziol123, line 114: token color has bigger priority than hex

lib/js/components/RichList/RichListGroupItem/RichListGroupItem.vue
```diff
@@ -0,0 +1,122 @@
+<template>
+	<div
+		class="richListGroupItem"
+		:class="{
+			[borderColorClass]: !!borderColor,
+		}"
+		:style="{
+			borderColor: borderColorHex ? borderColorHex : null,
```
pioziol123, line 8: It is not this border color we should change :)

lib/js/components/RichList/RichListGroupItem/RichListGroupItem.vue
```diff
@@ -0,0 +1,122 @@
...
+import { PropType } from 'vue';
+import {
+	RICH_LIST_GROUP_ITEM_BORDER_COLOR,
+	RichListGroupItemBorderColor,
+} from './RichListGroupItem.consts';
+
+export default {
+	name: 'RichListGroupItem',
+	components: {},
+	props: {
```
pioziol123, line 95: So I think about if all the other props are needed, and probably they are not. But should we construct a story the way we can, with one click, change ex `isDimmed` prop of all children and the parent?

lib/js/components/RichList/RichListGroupItem/RichListGroupItem.vue
```diff
@@ -0,0 +1,122 @@
...
+			[borderColorClass]: !!borderColor,
+		}"
+		:style="{
+			borderColor: borderColorHex ? borderColorHex : null,
+		}"
+	>
+		<div class="richListGroupItem__header">
+			<slot name="header" />
+		</div>
+		<div v-if="isExpanded" class="richListGroupItem__expanded">
```
pioziol123, line 14: class is unused

lib/js/components/RichList/RichListGroupItem/RichListGroupItem.vue
```diff
@@ -0,0 +1,122 @@
...
+		<div class="richListGroupItem__border" />
+	</div>
+</template>
+
+<style scoped lang="scss">
+@import '../../../../styles/settings/colors/tokens';
+@import '../../../../styles/settings/radiuses';
+@import '../../../../styles/settings/spacings';
+
+$rich-list-group-item-border-colors: (
```
pioziol123, line 26: Can we extract it somewhere? I bet it is the same list as for `RichListItem`

lib/js/components/RichList/RichListGroupItem/RichListGroupItem.vue
tweetgeek: I thought about it, but if I add a prop in this component that overwrites others, it may be confusing that the group also has such a prop. That's why I rejected it.

pioziol123 (approved):

PR #338, opened 2024-02-13
IT-5590 pill new props and slot accessory

pioziol123:
- [x] Pill renamed to Chip
- [x] Added slot accessory for svg instead of icon
- [x] Added LabelUppercase prop
- [x] Added ColorHex with specific color scheme overriding Color prop
- [x] Added Radius prop
- [x] HasDelete prop renamed to IsRemovable


rogatty:
It's a breaking change so remember to release as a v19. 🙂

lib/js/components/Outline/OutlineItem/OutlineItem.stories.ts
```diff
@@ -1,5 +1,5 @@
 import OutlineItem from './OutlineItem.vue';
-import DsPill from '../..//Pill/Pill.vue';
+import DsChip from '../..//Chip/Chip.vue';
```
tweetgeek, line 2:
```suggestion
import DsChip from '../../Chip/Chip.vue';
```

lib/js/components/Chip/Chip.stories.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export const Interactive = StoryTemplate.bind({});
+const args = {
+	label: 'Chip z labelem',
+	isLabelUppercase: false,
+	leftIcon: null,
+	isRemovable: false,
+	size: CHIP_SIZES.SMALL,
+	color: CHIP_COLORS.NEUTRAL,
+	colorHex: null,
```
tweetgeek, line 36: add it to argTypes, without on start we got red border in storybook actions

lib/js/components/Chip/Chip.stories.ts
```diff
@@ -0,0 +1,71 @@
...
+	data() {
+		return {
+			ICONS: Object.freeze(ICONS),
+		};
+	},
+});
+
+export const Interactive = StoryTemplate.bind({});
+const args = {
+	label: 'Chip z labelem',
```
tweetgeek, line 30: order of args is not the same as in miro

lib/styles/settings/_radiuses.scss
```diff
@@ -2,3 +2,5 @@ $radius-xs: 2px;
 $radius-s: 4px;
 $radius-m: 6px;
 $radius-l: 8px;
+//Special radius for components which have to be always rounded
+$radius-xl: 200px;
```
tweetgeek, line 6: should be as `xl`? maybe more specific? what if we want to add a border, e.g. 12 px as xl?

lib/js/components/Chip/Chip.vue
```diff
@@ -120,12 +126,20 @@ $pill-colors: (
 			'background': $color-default-background,
 		),
 	),
+	'invertedHex': (
```
tweetgeek, line 129: the name is a bit misleading because it is not inverted hex

lib/styles/settings/_radiuses.scss
pioziol123: https://bethinkteam.slack.com/archives/C012R8RM3A4/p1707753780620249?thread_ts=1707753172.775679&cid=C012R8RM3A4
It is the same as in figma

lib/js/components/Chip/Chip.vue
pioziol123: Any proposal?

tweetgeek: just inverted?

PR #342, opened 2024-03-14
IT-5682 DrawerHeader adjustments

fraszczakszymon:


lib/js/components/Drawer/DrawerHeader/DrawerHeader.vue
```diff
@@ -22,7 +22,7 @@
 				>
 					{{ eyebrowText }}
 				</span>
-				<div class="drawerHeader__title" :class="{ [`-${titleColor}`]: true }">
```
pioziol123, line 25: Too much buddy, too much

lib/js/components/Drawer/DrawerHeader/DrawerHeader.vue
```diff
@@ -129,6 +138,11 @@ $minimal-drawer-header-height: 82px;
 		padding: $space-m $space-xs $space-m $space-s;
 	}
 
+	&__actions {
+		align-self: flex-start;
```
pioziol123, line 142: Is misaligned with close icon

pioziol123: I think we can remove align-self here and remove __closeButton class, and it will be as in Figma

PR #343, opened 2024-03-14
Add 1600px breakpoint

gitkuba:


lib/styles/settings/_media-queries.scss
```diff
@@ -9,3 +9,7 @@
 @function breakpoint-l() {
 	@return '(min-width: 1280px)';
 }
+
+@function breakpoint-xl() {
```
kvas-damian, line 13: should we show it in https://github.com/bethinkpl/design-system/blob/442a99fffbc3cbb659d57288e1c44e9407c9f40b/.storybook/preview.js#L33 ?

lib/styles/settings/_media-queries.scss
```diff
@@ -9,3 +9,7 @@
...
+	@return '(min-width: 1600px)';
```
kvas-damian, line 14: Is it something that UX agreed on?

gitkuba: Yes.

kvas-damian (approved):

PR #345, opened 2024-03-15
IT-5546 Switch component

fraszczakszymon:


lib/js/components/Switch/Switch.vue
```diff
@@ -0,0 +1,350 @@
...
+		min-height: 24px;
+		min-width: 48px;
+		padding: 0 $space-xs;
+	}
+
+	&.-rounded &__item {
+		border-radius: $radius-xl;
+	}
+
+	&.-disabled &__item {
```
pioziol123, line 199: for disabled we usually also use `pointer-events: none;`

pioziol123: and let's group styles specific for $.-disabled. 
ex 
```
$.-disabled {
  #{$root}__icon {...}
  #{$roor}__label {...}
}
```

lib/js/components/Switch/Switch.vue
```diff
@@ -0,0 +1,350 @@
...
+	emits: ['update:selectedSide'],
+	data() {
+		return {
+			ICONS: Object.freeze(ICONS),
+			ICON_SIZES: Object.freeze(ICON_SIZES),
+			SWITCH_RADIUSES: Object.freeze(SWITCH_RADIUSES),
+			SWITCH_SIDE: Object.freeze(SWITCH_SIDE),
+			SWITCH_SIZE: Object.freeze(SWITCH_SIZES),
+			SWITCH_STATE: Object.freeze(SWITCH_STATE),
+			currentSide: this.selectedSide,
```
pioziol123, line 333: Let's not rewrite the side but always use side that comes from props.

lib/js/components/Switch/Switch.vue
```diff
@@ -0,0 +1,350 @@
...
+		padding: 0 $space-s;
+		position: relative;
+		transition: $switch-transition;
+		z-index: 2;
+
+		&.-left {
+			margin: -1px 0 -1px -1px;
+		}
+
+		&.right {
```
pioziol123, line 144:
```suggestion
		&.-right {
```

lib/js/components/Switch/Switch.vue
```diff
@@ -0,0 +1,350 @@
...
+
+	&.-disabled &__label {
+		color: $color-neutral-text-weak;
+	}
+
+	&.-disabled .-selected &__label {
+		color: $color-neutral-text-strong;
+	}
+
+	&:not(.-disabled) &__item:not(.-selected) {
```
pioziol123, line 224: Instead of this complicated selector, maybe we should name it and compute if the item should be attributed with it with JavaScript?

lib/js/components/Switch/Switch.vue
```diff
@@ -0,0 +1,350 @@
...
+		&:hover {
+			background-color: $color-neutral-background-ghost-hovered;
+
+			&::before {
+				background-color: $color-neutral-ripple;
```
pioziol123, line 229: I think ripple effect should be shared across all DS components, but I need to clarify it first.
https://bethinkteam.slack.com/archives/C018TDAH5FW/p1710777401086249

lib/js/components/Switch/Switch.vue
```diff
@@ -0,0 +1,350 @@
...
+			&.-right {
+				left: calc(100% + 2px);
+				transform: translateX(-100%);
+			}
+		}
+
+		&.-selected,
+		&.-selection:hover {
+			cursor: default;
+			flex-grow: 0.00001;
```
pioziol123, line 181: It looks like a hack. What it does?

lib/js/components/Switch/Switch.vue
```diff
@@ -0,0 +1,350 @@
...
+<style lang="scss" scoped>
+@import '../../../styles/settings/animations';
+@import '../../../styles/settings/spacings';
+@import '../../../styles/settings/radiuses';
+@import '../../../styles/settings/colors/tokens';
+@import '../../../styles/settings/typography/tokens';
+
+$switch-transition: all $default-transition-time ease-out;
+
+.switchWrapper {
```
pioziol123, line 78: Wy do we need a wrapper?

lib/js/components/Switch/Switch.vue
```diff
@@ -0,0 +1,350 @@
...
+				</div>
+			</div>
+			<div
+				class="switch__item -selection"
+				:class="{
+					'-left': currentSide === SWITCH_SIDE.LEFT,
+					'-right': currentSide === SWITCH_SIDE.RIGHT,
+				}"
+			>
+				<ds-icon
```
pioziol123, line 57: I see this is a hacky way to make the slider keep the same wide as a give option. Did you explore usage of grid mode? https://codepen.io/Piotr-Zi-kowski/pen/YzMNqoa

lib/js/components/Switch/Switch.vue
fraszczakszymon: To prevent overflowing options in case just one of them is too long for the parent container. It's an edge case. I get rid of it if you believe this is not going to be the case :+1:

lib/js/components/Switch/Switch.vue
fraszczakszymon: Yes, this is a workaround. Changing `flex-grow` from 1 to 0 does not trigger `transition` animation and changes element width without growing/shrinking effect. Having small, non-zero value gives us animation.

lib/js/components/Switch/Switch.vue
```diff
@@ -0,0 +1,350 @@
...
+		iconRight: {
+			type: Object as PropType<IconItem>,
+			default: null,
+			validator(icon: IconItem) {
+				return Object.values(ICONS).includes(toRaw(icon));
+			},
+		},
+		labelLeft: {
+			type: String,
+			default: null,
```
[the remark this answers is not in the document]
fraszczakszymon: Changed to `''`

lib/js/components/Switch/Switch.vue
```diff
@@ -0,0 +1,350 @@
...
+			currentIcon: this.selectedSide === SWITCH_SIDE.LEFT ? this.iconLeft : this.iconRight,
+			currentLabel: this.selectedSide === SWITCH_SIDE.LEFT ? this.labelLeft : this.labelRight,
+		};
+	},
+	methods: {
+		onSwitch(side: SwitchSelection) {
+			if (this.state === SWITCH_STATE.DISABLED || this.currentSide === side) {
+				return;
+			}
+			this.currentIcon = side === SWITCH_SIDE.LEFT ? this.iconLeft : this.iconRight;
```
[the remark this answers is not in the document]
fraszczakszymon: Absolutely! Thanks!

lib/js/components/Switch/Switch.vue
fraszczakszymon: I fully agree with that. However, there is a problem with storybook (known issue). Once you change value through the props (controls) the storybook iframe is being refreshed and it breaks the animation preview. Also, I don't like storybook-smelling code in the production component, so I can address it after testing. What do you think?

Alternatively, it looks like is has been addressed in the newer version of storybook. But upgrading DB doesn't seem to be in scope of this ticket.

lib/js/components/Switch/Switch.vue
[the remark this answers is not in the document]
pioziol123: Sorry for not being specific. I was thinking about making it a required prop. But I looked into miro and we mark required props explicitly so maybe it was fine 🙈

lib/js/components/Switch/Switch.vue
pioziol123: Let's leave it :)

lib/js/components/Switch/Switch.vue
pioziol123: We decided on daily that this is fine until we upgrade storybook to a newer version. Let's leave a comment here

lib/js/components/Switch/Switch.vue
```diff
@@ -0,0 +1,336 @@
...
+@import '../../../styles/settings/spacings';
+@import '../../../styles/settings/radiuses';
+@import '../../../styles/settings/colors/tokens';
+@import '../../../styles/settings/typography/tokens';
+
+$switch-transition: all $default-transition-time ease-out;
+
+.switchWrapper {
+	display: inline-flex;
+	max-width: 100%;
```
pioziol123, line 84: Does it anything? I cannot see any effect when i disable this proprety. Same for `max-width` below

lib/js/components/Switch/Switch.vue
```diff
@@ -0,0 +1,336 @@
...
+	overflow: hidden;
+}
+
+.switch {
+	$root: &;
+
+	align-items: flex-start;
```
pioziol123, line 91: I don't see any difference after removing this property

lib/js/components/Switch/Switch.vue
```diff
@@ -0,0 +1,336 @@
...
+	background-color: $color-default-background;
+	border: 1px solid $color-neutral-border;
+	border-radius: $radius-xs;
+	box-sizing: border-box;
```
pioziol123, line 95: I don't see any difference after removing this property

lib/js/components/Switch/Switch.vue
```diff
@@ -0,0 +1,336 @@
...
+
+		&.-right {
+			margin: -1px -1px -1px 0;
+		}
+
+		&.-selection {
+			background-color: $color-primary-background;
+			border: 1px solid $color-primary-border;
+			box-sizing: border-box;
+			left: 0;
```
pioziol123, line 156: It is unnecessary

lib/js/components/Switch/Switch.vue
```diff
@@ -0,0 +1,336 @@
...
+			&.-right {
+				left: calc(100% + 1px);
+				transform: translateX(-100%);
+			}
+		}
+
+		&.-selected,
+		&.-selection:hover {
+			cursor: default;
+			flex-shrink: 0;
```
pioziol123, line 185: What the flex-shrink does?

lib/js/components/Switch/Switch.vue
```diff
@@ -0,0 +1,336 @@
...
+			overflow: hidden;
```
pioziol123, line 186: Is it necessary? It is already in __item class

lib/js/components/Switch/Switch.vue
```diff
@@ -0,0 +1,336 @@
...
+			text-overflow: initial;
+		}
+
+		&.-clickable:hover {
+			background-color: $color-neutral-background-ghost-hovered;
+		}
+	}
+
+	&.-small &__item {
```
pioziol123, line 195: Maybe it should be an `__item` internal modifier class since it only modifies its elements

lib/js/components/Switch/Switch.vue
```diff
@@ -0,0 +1,336 @@
...
+		gap: $space-xxxxs;
+		min-height: 24px;
+		min-width: 48px;
+		padding: 0 $space-xs;
+	}
+
+	&.-rounded &__item {
```
pioziol123, line 202: Please put it under `.-rounded` above.  It is the way we do everywhere, and is easier, at least to me, to read such code.

lib/js/components/Switch/Switch.vue
fraszczakszymon: Limits wrapper to max width of the container. Effect without this line:
image

lib/js/components/Switch/Switch.vue
```diff
@@ -0,0 +1,336 @@
...
+
+	&.-rounded {
+		border-radius: $radius-xl;
+	}
+
+	&__icon {
+		color: $color-neutral-icon;
+	}
+
+	.-selected &__icon {
```
pioziol123, line 108: Please group this selector with the one below

lib/js/components/Switch/Switch.vue
pioziol123: Oh. That's a new thing to me :P thx

lib/js/components/Switch/Switch.vue
pioziol123: Someone would have to forcefully overwrite `.switchWrapper` class styles, no? Then I think it is not necessary. We don't overwrite components styles, it is a no no :P

lib/js/components/Switch/Switch.vue
fraszczakszymon: Then it needs to be defined in 3 places: `.item.-left`, `.item.-right` and `.selection`. I wasn't sure if it's better to avoid duplication or simplify the css selector here. I'm fine with both approaches.

pioziol123: Ok. So leave it here then

PR #346, opened 2024-03-18
IT-5545 Update SectionHeader and use it in DrawerSection

fraszczakszymon:


lib/js/components/Headers/SectionHeader/SectionHeader.vue
```diff
@@ -238,6 +310,38 @@
...
+		#{$root}__titleWrapper {
+			gap: $space-xxxs;
+		}
+
+		#{$root}__header {
+			@include info-s-extensive-bold-uppercase;
+		}
+
+		#{$root}__icon {
+			color: $color-neutral-icon-weak;
```
pioziol123, line 327: should the color be overwritten here?

lib/js/components/Headers/SectionHeader/SectionHeader.vue
```diff
@@ -79,9 +85,75 @@
...
+		'colorHover': $color-warning-icon-hovered,
+	),
+);
+
 .sectionHeader {
 	$root: &;
 
+	@each $color-name, $color-map in $icon-colors {
+		&__icon.-#{$color-name},
+		&.-size-xx-small &__icon.-#{$color-name} {
```
pioziol123, line 148: just __icon isn't enough

lib/js/components/Headers/SectionHeader/SectionHeader.vue
fraszczakszymon: Yes, xx-small has weaker color by default

lib/js/components/Headers/SectionHeader/SectionHeader.vue
fraszczakszymon: Nope, because of different default color for xx-small

lib/js/components/Drawer/DrawerSection/DrawerSection.stories.ts
```diff
@@ -12,25 +18,97 @@ const StoryTemplate: StoryFn<typeof DrawerSection> = (args) => ({
...
+			Default slot
+			<template #uncollapsible>Uncollapsible slot</template>
 		</drawer-section>`,
 });
 
 export const Interactive = StoryTemplate.bind({});
 
 const args = {
-	hasDivider: true,
+	size: SECTION_HEADER_SIZES.XX_SMALL,
```
pioziol123, line 57: I wonder if this should be accessible. In figma we have that xxs size should be used 
https://www.figma.com/file/izQdYyiBR1GQgFkaOIfIJI/LMS---DS-Components?type=design&node-id=6785-111048&mode=dev

lib/js/components/Drawer/DrawerSection/DrawerSection.stories.ts
```diff
@@ -12,25 +18,97 @@ const StoryTemplate: StoryFn<typeof DrawerSection> = (args) => ({
...
+			:size="size"
 			:title="title"
+			:supporting-text="supportingText"
+			:eyebrow="eyebrow"
+			:icon-left="ICONS[iconLeft]"
+			:icon-left-color="iconLeftColor"
+			:icon-right="ICONS[iconRight]"
+			:icon-right-color="iconRightColor"
+			:info="info"
+			:divider="divider"
```
pioziol123, line 42: Let's rename it in SectionHeader to`hasDivider`

lib/js/components/Drawer/DrawerSection/DrawerSection.vue
```diff
@@ -1,106 +1,137 @@
...
+			type: String,
+			default: SECTION_HEADER_MOBILE_LAYOUTS.VERTICAL,
+			validator: (value) => Object.values(SECTION_HEADER_MOBILE_LAYOUTS).includes(value),
+		},
 	},
+	emits: ['info-click', 'update:isExpanded'],
 	data() {
 		return {
-			isExpanded: true,
+			isSectionHeaderExpanded: this.isExpanded,
```
pioziol123, line 122: Let's leave a comment to remove this additional variable when upgrading do new Storybook version

lib/js/components/Drawer/DrawerSection/DrawerSection.stories.ts
fraszczakszymon: It's listed as one of prop in miro:
image

pioziol123: Yes, and it is inconsistent with Figma

lib/js/components/Drawer/DrawerSection/DrawerSection.stories.ts
pioziol123: Some time ago we started to add those `has` and `is` prefixes. It would be nice to make it consistent and aligned with current way of naming props. It is a suggestion, tho

lib/js/components/Drawer/DrawerSection/DrawerSection.vue
fraszczakszymon: Thanks for reminding! I removed it here as it's not really needed. The comment should be added in the Switch component: https://github.com/bethinkpl/design-system/pull/347

lib/js/components/Headers/SectionHeader/SectionHeader.vue
```diff
@@ -377,6 +446,16 @@ export default {
 			return ICON_SIZES.XX_SMALL;
 		},
 	},
+	watch: {
+		isExpanded: {
```
pioziol123, line 450: This property isn't reactive?

PR #350, opened 2024-03-21
IT-5546 Update internal state on prop change in switch

fraszczakszymon:


lib/js/components/Switch/Switch.vue
```diff
@@ -316,6 +316,16 @@ export default {
 			return this.currentSide === SWITCH_SIDE.LEFT ? this.labelLeft : this.labelRight;
 		},
 	},
+	watch: {
+		selectedSide: {
```
tweetgeek, line 320: I analyzed the entire switch code, maybe it was a specific decision. But why do we use the `currentSide` variable? Wouldn't it be better to base it on `selectedSide` prop? Then we don't need this watch at all.

tweetgeek (approved):

PR #360, opened 2024-05-07
LMS-1855 | add none to POP_OVER_TRIGGER_ACTIONS

preziu:


lib/js/components/PopOver/PopOver.vue
```diff
@@ -12,32 +12,34 @@
...
-				'-medium': size === POP_OVER_SIZES.MEDIUM,
-			}"
-		>
-			<img v-if="headerImageUrl" class="popOver__image" :src="headerImageUrl" alt="" />
-			<div class="popOver__content">
-				<div v-if="titleText" class="popOver__title"> {{ titleText }} </div>
-				<div v-if="subtitleText" class="popOver__subtitle"> {{ subtitleText }} </div>
-				<div class="popOver__contentSlot" :class="{ '-maxHeight': maxHeight }">
-					<slot :close="close" />
+		<template v-if="shouldDisplay">
```
tweetgeek, line 15: I think it should be a level higher. We shouldn't render vue-popper

lib/js/components/PopOver/PopOver.vue
```diff
@@ -272,6 +274,10 @@ export default {
 			type: String,
 			default: '',
 		},
+		shouldDisplay: {
+			type: Boolean,
+			default: true,
+		},
```
tweetgeek, line 280:
```suggestion
		disabled: {
			type: Boolean,
			default: false,
		},
```

tweetgeek: or maybe we should add a new `POP_OVER_TRIGGER_ACTIONS` e.g. `none`?

preziu: @tweetgeek using `none` is much more cleaner, please check again

lib/js/components/PopOver/PopOver.consts.ts
```diff
@@ -6,6 +6,7 @@ export const POP_OVER_COLORS = {
 export const POP_OVER_TRIGGER_ACTIONS = {
 	CLICK: 'click',
 	HOVER: 'hover',
+	NONE: 'none',
```
tweetgeek, line 9: are you sure that's enough? If we set `none`, we still render the html of this popover. Is this a good approach? In a component such as lesson status, we will always render the content of the popover, which will be unnecessary for most users - because they only see this popover once during onboarding.
Zrzut ekranu 2024-05-9 o 11 54 10

tweetgeek (approved):

PR #369, opened 2024-05-27
Use plugin:vue/vue3-recommended for eslint

rogatty:
We were using rules for Vue 2, my bad.

I clicked through all of the involved stories and they all seem fine to me. We're not using `update:isSelected` anywhere in the platform so that rename shouldn't be an issue.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,7 +2,7 @@ module.exports = {
 	plugins: ['import', '@typescript-eslint'],
 	extends: [
 		'eslint:recommended',
-		'plugin:vue/recommended',
+		'plugin:vue/vue3-recommended',
 		'prettier',
 		'plugin:storybook/recommended',
 	],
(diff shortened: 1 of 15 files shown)


.eslintrc.js
```diff
@@ -126,14 +126,5 @@ module.exports = {
...
 		'vue/require-emit-validator': 'error',
-
-		// Required for Vue 3 migration
-		'vue/no-deprecated-slot-attribute': 'error',
-		'vue/no-deprecated-slot-scope-attribute': 'error',
-		'vue/require-explicit-emits': 'error',
-
-		// https://v3-migration.vuejs.org/breaking-changes/key-attribute.html
-		'vue/no-v-for-template-key': ['off'],
-		'vue/no-v-for-template-key-on-child': 'error',
```
rogatty, line 137: These are a part of `plugin:vue/vue3-recommended`

PR #375, opened 2024-06-18
Update build-dist-on-trigger.yml to post message in PR with install details

tweetgeek:


.github/workflows/build-dist-on-trigger.yml
```diff
@@ -56,3 +56,18 @@ jobs:
           commit_message: Build dist
           create_branch: true
           push_options: '--force'
+
+      - name: Get last commit hash from branch and set to variable
+        id: last-commit
+        run: |
+          git fetch origin ${{ steps.comment-pr-ref.outputs.result }}-dist
```
rogatty, line 63: We already have this commit checked out at the end of the previous step. There is probably a way to get its commit hash without fetching from origin. I don't think it's a blocker though, just an optimisation for the future.

rogatty (approved):

PR #376, opened 2024-06-24
IT-5798 Add FA_SIDEBAR_FLIP icon

fraszczakszymon:


lib/js/icons/fontawesome.ts
```diff
@@ -338,6 +339,7 @@ export const FONTAWESOME_ICONS = {
 	FA_SHARE: faShare,
 	FA_SHARE_NODES: faShareNodes,
 	FA_SHIELD_HALVED: faShieldHalved,
+	FA_SIDEBAR_FLIP: faSidebarFlip,
```
kvas-damian, line 342: do we have to use flipped version? We could flip the icon in CSS or via props on Icon. This way, if we find a use-case for regular `SIDEBAR` icon, we won't need to import 2 separate assets

PR #382, opened 2024-07-02
IT-5852 Upgrade storybook and start with vite

fraszczakszymon:


vite.config.ts
```diff
@@ -0,0 +1,42 @@
+import { resolve } from 'path';
+import { defineConfig } from 'vite';
+import vue from '@vitejs/plugin-vue';
+import checker from 'vite-plugin-checker';
+import svgLoader from 'vite-svg-loader';
+import dts from 'vite-plugin-dts';
+
+// https://vitejs.dev/config/
+export default defineConfig({
```
fraszczakszymon, line 9: Note to keep in mind recent changes: https://github.com/bethinkpl/design-system/pull/404/files#diff-1d62a367f02f5c91000e35c82b0086fae3246eca3942f9034c058b16a3780ec2R22

.storybook/main.ts
```diff
@@ -0,0 +1,71 @@
...
+				buildStart() {
+					if (isProductionMode()) {
+						const result = sass.compile(previewStylesPath);
+
+						if (!fs.existsSync(path.resolve(__dirname, '../public/storybook'))) {
+							fs.mkdirSync(path.resolve(__dirname, '../public/storybook'), { recursive: true });
+						}
+
+						fs.writeFileSync(cssFilePath, result.css.toString(),);
+					}
```
rogatty, line 43: Let's leave an explanation here, why this is needed.

.storybook/preview.scss
```diff
@@ -0,0 +1,3 @@
+@import '../lib/styles/normalize';
+@import '../lib/styles/design-system';
+@import '../lib/styles/storybook';
```
rogatty, line 3: How about moving this file to `/.storybook`? Would that work? It's not a part of the exported lib, so the current path stinks a bit.

package.json
rogatty, line 1: I'm getting this warning when running `yarn storybook`:
```
The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.
```
Adding `"type": "module"` seems to help but I'm not yet sure if it breaks anything.

lib/js/typings.d.ts
```diff
@@ -1,8 +1,9 @@
 // TypeScript compiler doesn't read webpack config so we need to tell him how interpret import Vue files
```
rogatty, line 2:
```suggestion
```

package.json
```diff
@@ -11,12 +11,12 @@
...
   "license": "UNLICENSED",
   "engines": {
-    "yarn": "^1.22.0",
+    "yarn": "^1.22.1",
     "node": "^16.13.0 || ^18.17.0 || ^20.9.0"
   },
   "scripts": {
-    "build": "vue-cli-service build --target lib --formats umd --name design-system lib/js/index.ts",
-    "dev": "start-storybook -p 6006",
+    "build": "vite build",
```
rogatty, line 18: Previously we were building only umd, now we're building two formats:
```
dist/RadioButton-BSIC6h6x.mjs    1.76 kB │ gzip:   0.83 kB │ map:     3.92 kB
dist/pl-B2O3Uvpn.mjs             2.13 kB │ gzip:   0.96 kB │ map:     3.26 kB
dist/index-Oq81NaeJ.mjs         64.85 kB │ gzip:  16.54 kB │ map:   180.53 kB
dist/design-system.mjs         237.31 kB │ gzip:  25.77 kB │ map:     0.10 kB
dist/index-C_3Wau9G.mjs        771.47 kB │ gzip: 157.69 kB │ map: 4,713.17 kB
dist/design-system.umd.js  894.01 kB │ gzip: 179.64 kB │ map: 4,821.77 kB
```

How about building only one of them, whichever works best?

package.json
```diff
@@ -74,18 +77,22 @@
...
     "ts-jest": "29.0.5",
     "ts-node": "^10.9.1",
     "typescript": "4.6.4",
+    "vite": "^5.4.7",
+    "vite-plugin-checker": "^0.7.0",
+    "vite-plugin-css-injected-by-js": "^3.5.1",
+    "vite-plugin-dts": "^3.9.1",
+    "vite-svg-loader": "^5.1.0",
     "vue": "3.2.45",
     "vue-loader": "17.0.0",
```
rogatty, line 94: I think we can remove it together with `@vue/cli-*`, `sass-loader`, `vite.config.js` and maybe something more.

vite.config.ts
```diff
@@ -0,0 +1,37 @@
...
+	plugins: [
+		vue(),
+		checker({
+			typescript: true,
+		}),
+		svgLoader(),
+		dts(),
+		cssInjectedByJsPlugin(),
```
rogatty, line 18: Let's explain in a comment why we need this.

lib/js/components/Form/Checkbox/Checkbox.stories.ts
```diff
@@ -3,32 +3,37 @@ import Checkbox from './Checkbox.vue';
...
 
 export const Interactive = StoryTemplate.bind({});
 
 Interactive.argTypes = argTypes(CHECKBOX_SIZE, CHECKBOX_STATE);
 
-Interactive.args = args;
+Interactive.args = args(CHECKBOX_SIZE, CHECKBOX_STATE);
 
 Interactive.parameters = {
 	actions: {
```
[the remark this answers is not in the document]
rogatty: Maybe because of the missing decorator? https://storybook.js.org/docs/essentials/actions#action-event-handlers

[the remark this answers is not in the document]
rogatty: I'm not pushing to fix it now as it never worked perfectly and is for debugging only. Maybe https://github.com/bethinkpl/design-system/blob/c00b4f8cfd1df903e826bff70821765362d024fe/lib/js/components/Headers/OverlayHeader/OverlayHeader.stories.ts#L17-L21 is a better approach and we should migrate at some point.

tsconfig.json
```diff
@@ -20,21 +21,27 @@
...
+        "lib/*"
+      ]
     },
     "declaration": true,
-    "declarationDir": "dist"
+    "declarationDir": "dist",
+    "types": ["node", "jest"],
   },
   "include": [
+    "vite.config.*",
```
rogatty, line 33: `vite.config.ts`?

vite.config.ts
rogatty: @fraszczakszymon what about it? Do we need to fix anything here?

lib/js/primevue.ts
```diff
@@ -1,7 +1,10 @@
 import PrimeVue from 'primevue/config';
+import { definePreset } from '@primeuix/styled';
 import Aura from '@primevue/themes/aura';
 import Tooltip from 'primevue/tooltip';
 
+const BethinkPreset = definePreset({});
```
rogatty, line 6: Seems unused, does it make sense to keep it?

vite.config.ts
fraszczakszymon: Tested. It's fine. I forgot to resolve this comment 👍🏻

lib/js/components/Drawer/DrawerSection/DrawerSection.vue
```diff
@@ -21,7 +21,9 @@
 		<div v-if="isExpandedInternal || !isExpandable">
 			<slot />
 		</div>
-		<div v-if="$slots.uncollapsible"><slot name="uncollapsible" /></div>
+		<div v-if="$slots.uncollapsible">
+			<slot name="uncollapsible" />
+		</div>
```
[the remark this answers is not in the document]
fraszczakszymon: It does not come from prettier 🤔 reverted

vite.config.ts
fraszczakszymon: Isn't it self explanatory? Webpack used to build in css into js by default, vite requires additional plugin for that. We rely on that - we don't import component styles separately in platform

deploy_storybook_to_s3.sh
```diff
@@ -19,6 +20,20 @@ echo "I am building a storybook"
...
+    echo "s3://$BUCKET/$BRANCH/$RELATIVE_PATH"
+    s3cmd modify --add-header="Content-Type: application/javascript" s3://$BUCKET/$BRANCH/$RELATIVE_PATH
+done
+
+echo "Setting Content-Type to text/css for all .css files"
+find "$LOCAL_DIRECTORY" -type f -name "*.css" | while read FILE; do
+    RELATIVE_PATH="${FILE#"$LOCAL_DIRECTORY/"}"
+    echo "s3://$BUCKET/$BRANCH/$RELATIVE_PATH"
+    s3cmd modify --add-header="Content-Type: text/css" s3://$BUCKET/$BRANCH/$RELATIVE_PATH
+done
```
rogatty, line 37: There are so many assets and it takes so long... I hope we find a better solution soon.

rogatty (approved):

PR #389, opened 2024-07-16
IT-5827 | Add Image component

rogatty:
https://primevue.org/skeleton/
https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit


rogatty:
It's a part of https://github.com/bethinkpl/design-system/pull/391 now.

PR #391, opened 2024-07-22
IT-5821 | Add media slot to RichListItem + Image and Skeleton components

rogatty:
DR: https://bethinkteam.slack.com/archives/CFBLF5RGA/p1721652437934089


lib/js/components/Image/Image.vue
```diff
@@ -0,0 +1,79 @@
...
+			:class="{
+				'-ds-contain': fit === IMAGE_FITS.CONTAIN,
+				'-ds-cover': fit === IMAGE_FITS.COVER,
+			}"
+			loading="lazy"
+			:src="src"
+			@error="isLoading = false"
+			@load="isLoading = false"
+		/>
+		<div v-if="isLoading" class="ds-image__loader">
```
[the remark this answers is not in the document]
rogatty: Prime adds `position: relative` to inline styles in https://github.com/primefaces/primevue/blob/350cb5805a7150ded6029a79a357055c25125445/packages/primevue/src/skeleton/style/SkeletonStyle.js#L41-L43 and we need `position: absolute`. In theory we could play with overrides but it smells and may break at some point. I'd say it's not worth the trouble.

lib/js/components/Image/Image.vue
```diff
@@ -0,0 +1,81 @@
...
+				'-ds-cover': fit === IMAGE_FITS.COVER,
+			}"
+			draggable="false"
+			loading="lazy"
+			:src="src"
+			@error="isLoading = false"
+			@load="isLoading = false"
+		/>
+		<div v-if="isLoading" class="ds-image__loader">
+			<prime-skeleton width="100%" height="100%" border-radius="0" />
```
rogatty, line 16: https://primevue.org/skeleton/

lib/js/components/Image/Image.vue
```diff
@@ -0,0 +1,81 @@
...
+		user-select: none;
+		width: 100%;
+
+		&.-ds-contain {
+			object-fit: contain;
+		}
+
+		&.-ds-cover {
+			object-fit: cover;
+		}
```
rogatty, line 37: https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit

lib/js/components/Image/Image.vue
```diff
@@ -0,0 +1,81 @@
...
+import { IMAGE_FITS, ImageFit } from './Image.consts';
+
+export default defineComponent({
+	// <image> is "an ancient and poorly supported precursor to the <img> element"
+	// so there should be no conflicts. Also, we prefix all DS components with ds-
+	// https://developer.mozilla.org/en-US/docs/Web/HTML/Element#obsolete_and_deprecated_elements
+	// eslint-disable-next-line vue/no-reserved-component-names
+	name: 'Image',
+	components: {
+		PrimeSkeleton,
```
fraszczakszymon, line 62: @rogatty Shouldn't we create DsSkeleton and use it here instead of using `primevue/skeleton` directly?

rogatty: I don't know, it might be a question for #design-system. We need to add a new component for wnl-platform but maybe we don't need it for DS components?

kvas-damian: IMO, we should have a DsSkeleton component and we should use it here. If we decide to introduce some changes (eg. colors or animation), this place should also follow that change.

lib/js/components/Image/Image.vue
```diff
@@ -0,0 +1,94 @@
...
+	// https://developer.mozilla.org/en-US/docs/Web/HTML/Element#obsolete_and_deprecated_elements
+	// eslint-disable-next-line vue/no-reserved-component-names
+	name: 'Image',
+	components: {
+		DsSkeleton,
+	},
+	props: {
+		// Use when there are performance issues when other elements are overlayed
+		// on the image, e.g. a RichListItem border covers a part of the image
+		accelerate: {
```
kvas-damian, line 73: why do we expose it as a prop? Doe w plan to disable it anywhere?

lib/js/components/Skeleton/Skeleton.vue
```diff
@@ -0,0 +1,49 @@
+<template>
+	<prime-skeleton
+		width="100%"
+		height="100%"
```
kvas-damian, line 4: why we don't expose `width` and `height` as props?

rogatty: We try to avoid hardcoded sizes, so having skeleton fill the available place sounds like the cleanest approach. Parent can use flexbox etc. to have it sized appropriately. We can add props when there is an actual need.

kvas-damian: Zrzut ekranu 2024-07-26 o 08 14 15

IMO, there is an acttual need for this prop in OverlayHeader. Instead of creating a <div> wrapper for every line I could set `width="30%"` on a first line.

lib/js/components/Image/Image.vue
kvas-damian: If you listed so many downsides, should it be enabled by default?

BTW. strategy of exposing this prop "for future" is inconsistent with not exposing `width/height` props in Skeleton: https://github.com/bethinkpl/design-system/pull/391#discussion_r1691411648

rogatty: It shouldn't be enabled by default, my mistake.

lib/js/components/Skeleton/Skeleton.consts.ts
```diff
@@ -0,0 +1,7 @@
+export const SKELETON_RADIUSES = {
```
kvas-damian, line 1: what would you say about renaming it to `SKELETON_RADIUS_SIZES`?

kvas-damian (approved):

lib/js/components/Skeleton/Skeleton.consts.ts
rogatty: Sure, but let's wait for Ewa in https://bethinkteam.slack.com/archives/CFBLF5RGA/p1721995047212719?thread_ts=1721652437.934089&cid=CFBLF5RGA, in case she wants to change something else about radiuses.

lib/js/components/Skeleton/Skeleton.stories.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export const Interactive = StoryTemplate.bind({});
+
+Interactive.args = {
+	height: '100%',
+	radius: SKELETON_RADIUSES.MD,
+	width: '100%',
+} as Args;
+
+Interactive.argTypes = {
```
kvas-damian, line 29: please add link to Figma

PR #392, opened 2024-07-23
IT-5875 Update icons for MediaManager

fraszczakszymon:
Add:
- SIDEBAR_FLIP_SOLID
- FA_SIDEBAR_FLIP
- FA_SQUARE_LIST_SOLID

Remove:
- FA_SIDEBAR


rogatty:
@fraszczakszymon I see on Figma that we need more icons, both regular and solid versions.
Screenshot 2024-07-23 at 13 29 02

IMO we should wait with it until one of the styling PRs and you could use `FA_SIDEBAR` for now.

rogatty:
And we can remove `FA_SIDEBAR`.

lib/js/icons/fontawesome.ts
```diff
@@ -347,7 +348,7 @@ export const FONTAWESOME_ICONS = {
 	FA_SHARE: faShare,
 	FA_SHARE_NODES: faShareNodes,
 	FA_SHIELD_HALVED: faShieldHalved,
-	FA_SIDEBAR: faSidebar,
+	FA_SIDEBAR_FLIP: faSidebarFlip,
```
kvas-damian, line 351: Do we need to import flipped icons? Are we going to import flipped and non flipped version of this asset when we find a use-case for for the latter version? Why we don't use `FA_SIDEBAR` with `flippedHorizontal` prop?

rogatty: `IconButton` doesn't have `flippedHorizontal` prop.

kvas-damian: maybe it's time to expose it

rogatty: I'm cool with that but you should probably talk with UX. 🙂

kvas-damian: https://bethinkteam.slack.com/archives/CFBLF5RGA/p1721897628795609

kvas-damian (approved):

PR #401, opened 2024-08-06
IT-5906 DsTextGroup

fraszczakszymon:


lib/js/components/TextGroup/TextGroup.vue
```diff
@@ -0,0 +1,288 @@
...
+				'-ds-ellipsis': supportingTextEllipsis,
+			}"
+		>
+			<ds-skeleton
+				v-if="isLoading"
+				class="ds-textGroup__loadingSkeleton"
+				width="100%"
+				height="14px"
+			/>
+			<span v-else>{{ supportingText }}</span>
```
kvas-damian, line 63: In https://github.com/bethinkpl/wnl-platform/pull/5124#discussion_r1706897380 we're going to pass HTML into this prop. How do we handle HTML tags then? Escaped HTML doesn't look ok

fraszczakszymon: https://github.com/bethinkpl/design-system/pull/401/commits/a4f7027f06791412b682628f894ba785d4a56cd2

lib/js/components/TextGroup/TextGroup.vue
```diff
@@ -0,0 +1,296 @@
...
+	>
+		<div
+			v-if="eyebrowText !== null"
+			class="ds-textGroup__eyebrow"
+			:class="{
+				'-ds-ellipsis': eyebrowTextEllipsis,
+				'-ds-uppercase': isEyebrowTextUppercase,
+			}"
+		>
+			<ds-skeleton v-if="isLoading" width="50%" height="100%" />
```
kvas-damian, line 27: What if `eyebrowText` is constant, but we need to wait for `mainText` to be loaded? Should we display a skeleton in such case or eyebrow should be displayed?

lib/js/components/TextGroup/TextGroup.vue
```diff
@@ -0,0 +1,296 @@
...
+	$self: &;
+
+	display: flex;
+	flex-direction: column;
+	width: 100%;
+
+	&__eyebrow {
+		@include info-s-default-bold;
+
+		color: $color-neutral-text-weak;
```
kvas-damian, line 68: Optional: we could already prepare a map of color for `color` prop, so we don't need to analyze the component again. I suggest similar approach to: https://github.com/bethinkpl/design-system/blob/48818e1ab8537c238ee314967e38fd246f41d3af/lib/js/components/BadgeScore/BadgeScore.vue#L31

lib/js/components/TextGroup/TextGroup.vue
```diff
@@ -0,0 +1,296 @@
...
+		text-overflow: ellipsis;
+		white-space: nowrap;
+	}
+
+	&.-ds-small {
+		#{$self}__eyebrow {
+			@include info-xs-default-bold;
+
+			margin-bottom: 0;
+			min-height: 10px;
```
kvas-damian, line 104:
```suggestion
			min-height: $typography-line-height-5xs;
```
it isn't perfect, as variable name doesn't match typography name, but it's little bit safer than pixel value

lib/js/components/TextGroup/TextGroup.vue
```diff
@@ -0,0 +1,296 @@
...
+
+			&.-ds-uppercase {
+				@include info-xs-extensive-bold-uppercase;
+			}
+		}
+
+		#{$self}__main {
+			@include text-s-compact-bold;
+
+			min-height: 16px;
```
kvas-damian, line 114:
```suggestion
			min-height: $typography-line-height-2xs;
```

lib/js/components/TextGroup/TextGroup.vue
```diff
@@ -0,0 +1,296 @@
...
+		#{$self}__main {
+			color: $color-neutral-text-strong-disabled;
+		}
+
+		#{$self}__supporting {
+			color: $color-neutral-text-disabled;
+		}
+	}
+
+	&.-ds-disabled.-ds-selected.-ds-interactive:hover,
```
kvas-damian, line 148: usually, we set `pointer-events: none` on disabled elements, so there is no hover state. Is it different here?

lib/js/components/TextGroup/TextGroup.vue
kvas-damian: There is an alternative way to achieve similar result without these `min-heights`. We could use `&nbsp;` or another not visible char to reserve space in this container. Not sure if it's cleaner, but definitely easier to maintain the same height during style changes

lib/js/components/TextGroup/TextGroup.stories.ts
```diff
@@ -0,0 +1,111 @@
...
+	component: DsTextGroup,
+} as Meta<typeof DsTextGroup>;
+
+const StoryTemplate: StoryFn<typeof DsTextGroup> = (args) => ({
+	components: { DsDivider, DsTextGroup },
+	setup() {
+		return { ...args };
+	},
+	template: `
+		<div style="height: 200px; width: 280px;">
```
kvas-damian, line 22: showing this component on fixed width container makes it harder to understand how does it look like on big screens. Is it something UX wanted?

lib/js/components/TextGroup/TextGroup.vue
```diff
@@ -0,0 +1,296 @@
...
+	<div
+		class="ds-textGroup"
+		:class="{
+			'-ds-small': size === TEXT_GROUP_SIZES.SMALL,
+
+			'-ds-hovered': state === TEXT_GROUP_STATES.HOVERED,
+			'-ds-loading': isLoading,
+			'-ds-disabled': state === TEXT_GROUP_STATES.DISABLED,
+
+			'-ds-interactive': isInteractive,
```
kvas-damian, line 11: should we add `cursor: pointer` in `interactive` mode?

lib/js/components/TextGroup/TextGroup.vue
```diff
@@ -0,0 +1,296 @@
...
+			'-ds-selected': isSelected,
+
+			[loadingSizeClassName]: isLoading,
+
+			[colorClassName]: true,
+		}"
+	>
+		<div
```
kvas-damian, line 19: when `eyebrowtext:null` height of `loading` and `default` state of entire component doesn't match

kvas-damian: the same problem for other combinations of null text props

lib/js/components/TextGroup/TextGroup.vue
fraszczakszymon: https://bethinkteam.slack.com/archives/CFBLF5RGA/p1723103554903989

lib/js/components/TextGroup/TextGroup.vue
fraszczakszymon: Addressed using &nbsp;

lib/js/components/RichList/BasicRichListItem/BasicRichListItem.vue
```diff
@@ -71,43 +71,6 @@
...
-		}
-
-		&:hover {
-			color: $color-neutral-text-hovered;
-		}
-	}
-
 	&.-ds-small {
 		#{$root}__content {
 			padding: $space-3xs 0;
```
kvas-damian, line 76:
```suggestion
			padding: $space-2xs 0;
```
https://www.figma.com/design/izQdYyiBR1GQgFkaOIfIJI/LMS---DS-Components?node-id=8598-8165&m=dev

lib/js/components/TextGroup/TextGroup.vue
fraszczakszymon: https://github.com/bethinkpl/design-system/pull/401/commits/868b28ddd6ec59f2005867c8a763683b8c7ead24
+ 
https://bethinkteam.slack.com/archives/CFBLF5RGA/p1723109726737899?thread_ts=1723020458.153299&cid=CFBLF5RGA

kvas-damian (approved):

stylelint.config.js
```diff
@@ -44,6 +44,12 @@ module.exports = {
 		],
 		'order/properties-alphabetical-order': true,
 		'selector-class-pattern': '-?ds-[a-z][a-zA-Z_]*',
+		'unit-no-unknown': [
+			true,
+			{
+				ignoreUnits: ['lh'],
```
kvas-damian, line 50: is it ok, that we ignore it globally? Maybe it's better to disable error locally, so we can put a fallback there?

lib/js/components/TextGroup/TextGroup.vue
```diff
@@ -206,33 +212,18 @@ $text-group-colors: (
...
-		}
-
+		#{$self}__eyebrow,
+		#{$self}__main,
 		#{$self}__supporting {
-			height: 12px;
-			margin: 2px 0;
+			align-items: center;
+			display: flex;
+			height: 1.25em; // Fallback for older browsers
```
kvas-damian, line 220: if we could move this styles out of `.-ds-loading` then height won't be jumping on loading state on old browser - it'll be little off for all states

fraszczakszymon: Leaving as it is, otherwise we would need to apply this height only if ellipsis is enabled

kvas-damian (approved):

PR #405, opened 2024-08-21
IT-5925 | Show PopOver when forceShow is true and triggerAction is none

rogatty:
`forceShow: true` should always show the pop over, as the name suggests. Without it, there is no clean way to show feature onboarding using `triggerAction: none`.

https://bethinkpl.github.io/design-system/?path=/story/components-popover--interactive&args=placement:bottom-start;triggerAction:none;forceShow:true


lib/js/components/PopOver/PopOver.vue
rogatty, line 1: It works as expected but causes:
```
[Vue warn]: Invalid prop: custom validator check failed for prop "trigger".
```

rogatty:
We don't need it after all.

PR #413, opened 2024-09-02
IT-5903 Skip keyboard shortcuts on input elements

fraszczakszymon:


lib/js/components/Headers/OverlayHeader/OverlayHeader.vue
```diff
@@ -391,9 +392,12 @@ export default {
 		window.addEventListener('keydown', this.onKeydown);
 	},
 	methods: {
-		onKeydown(e) {
-			switch (e.keyCode) {
```
fraszczakszymon, line 395: keyCode is deprecated

lib/js/components/Headers/OverlayHeader/OverlayHeader.vue
```diff
@@ -391,9 +392,12 @@ export default {
...
-				case 81: // "Q" key
+		onKeydown(e: KeyboardEvent) {
+			if (isElementEditable(e.target as HTMLElement | null)) {
+				return;
+			}
+			switch (e.code) {
```
rogatty, line 399: Shouldn't we use `.key` instead?

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code
> For example, the code returned is "KeyQ" for the Q key on a QWERTY layout keyboard, but the same code value also represents the ' key on Dvorak keyboards and the A key on AZERTY keyboards. That makes it impossible to use the value of code to determine what the name of the key is to users if they're not using an anticipated keyboard layout.

> To determine what character corresponds with the key event, use the [KeyboardEvent.key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) property instead.

fraszczakszymon: I'm ok with `.key`, but added support for q and Q in order to catch [CAPSLOCK] + q and SHIFT + q.

rogatty (approved):

PR #416, opened 2024-09-06
IT-5961 remove top from ::before

pioziol123:


pioziol123:
don't know why but it seems to do the job. I don't see any visible changes in component

kvas-damian:
I break background on Chrome

PR #419, opened 2024-09-18
IT-5983 outline item slot label

pioziol123:


lib/js/components/Outline/OutlineItem/OutlineItem.vue
```diff
@@ -26,8 +26,15 @@
 				:icon="iconLeft"
 				:size="ICON_SIZES.X_SMALL"
 			/>
-			<span class="ds-outlineItem__text">
+			<span v-else class="ds-outlineItem__text">
```
kvas-damian, line 29: why `v-else`?

Isn't it icon + label a valid use-case?

lib/js/components/Outline/OutlineItem/OutlineItem.vue
```diff
@@ -26,8 +26,15 @@
...
 				<span class="ds-outlineItem__label" :class="{ '-ds-uppercase': isLabelUppercase }">
+					<template v-if="$slots.label">
+						<slot name="label" />
```
kvas-damian, line 32: should we mark it in a storybook that this component accept both prop and slot?

lib/js/components/Outline/OutlineItem/OutlineItem.vue
```diff
@@ -84,6 +91,7 @@
 		align-items: flex-start;
 		column-gap: $space-2xs;
 		display: flex;
+		max-width: 100%;
```
kvas-damian, line 94: what does it change?
When I put a long label without spaces, then text goes outside this component

pioziol123: Without it overflow does not work on .katex and .katex-display classes

lib/js/components/Outline/OutlineItem/OutlineItem.vue
pioziol123: yeah, u right 😒

kvas-damian (approved):

PR #424, opened 2024-10-11
IT-5385 date range picker

pioziol123:


lib/js/components/DatePickers/DatePickerBox/DatePickerBox.vue
```diff
@@ -242,6 +275,10 @@ export default defineComponent({
 			type: Date,
 			default: null,
 		},
+		endDate: {
```
pioziol123, line 278: I am not sure about naming and i am open for suggestions.

lib/js/components/DatePickers/DatePicker/DatePicker.composables.ts
pioziol123, line 1: I am not sure about how to name such files and where to put them, open for suggestions

lib/js/components/DatePickers/DatePicker/DatePicker.composables.ts
```diff
@@ -0,0 +1,87 @@
+import { onMounted, onUnmounted, Ref, ref, watch } from 'vue';
+import { Instance as DatePickerInstance } from 'flatpickr/dist/types/instance';
+import flatpickr from 'flatpickr';
```
pioziol123, line 3: I have no idea how to lazy load it

pioziol123: @rogatty what you mean about loading and error state?

lib/js/components/DatePickers/DatePicker/DatePicker.composables.ts
```diff
@@ -0,0 +1,108 @@
...
+	},
+	onChange: (dates: Array<Date>) => void,
+	defaultDates: Date | Array<Date>,
+	mode: 'single' | 'range' = 'single',
+): {
+	datePicker: DatePickerInstance | null;
+	createDatePicker: () => Promise<DatePickerInstance | undefined>;
+	isOpen: Ref<boolean>;
+	toggle: () => void;
+} {
```
tweetgeek, line 27:
```suggestion
interface InitFlatpickrPrams {
	flatpickrInputRef: Ref<HTMLInputElement>;
	dateRangePickerRef: Ref<HTMLElement>;
	props: {
		disableDates: Array<Date>;
		minDate: Date | null;
		maxDate: Date | null;
		calendarPosition: DatePickerCalendarPositions;
	};
	onChange: (dates: Array<Date>) => void;
	defaultDates: Date | Array<Date>;
	mode: 'single' | 'range' = 'single';
}

interface InitFlatpickr {
	datePicker: DatePickerInstance | null;
	createDatePicker: () => Promise<DatePickerInstance | undefined>;
	isOpen: Ref<boolean>;
	toggle: () => void;
}

export function initFlatpickr({
	flatpickrInputRef,
	dateRangePickerRef,
	props,
	onChange,
	defaultDates,
	mode = 'single',
}: InitFlatpickrPrams): InitFlatpickr {
```

Optional: in my opinion, using destructors and separate interfaces are more readable :)

lib/js/components/DatePickers/DatePicker/DatePicker.composables.ts
```diff
@@ -0,0 +1,108 @@
...
+	let datePicker: DatePickerInstance | null = null;
+	let isOpen = ref(false);
```
tweetgeek, line 40:
```suggestion
	const isOpen = ref(false);
```
strange that the linter does not catch this, but we do not overwrite `isOpen` anymore, we only change the value of the reference object - that's why `const`

lib/js/components/DatePickers/DatePicker/DatePicker.composables.ts
```diff
@@ -0,0 +1,108 @@
...
+
+	const createDatePicker = async () => {
```
tweetgeek, line 42:
```suggestion
	const createDatePicker = async (): Promise<FlatpickrFn> => {
```

lib/js/components/DatePickers/DatePicker/DatePicker.composables.ts
```diff
@@ -0,0 +1,108 @@
...
+		if (datePicker) {
+			return;
+		}
+		if (!flatpickrFunction) {
+			try {
+				flatpickrFunction = (await import('flatpickr')).default;
+				const { Polish } = await import('flatpickr/dist/l10n/pl');
+				locale = Polish;
```
tweetgeek, line 50: Optional:
```suggestion
				locale = (await import('flatpickr/dist/l10n/pl')).Polish;
```

lib/js/components/DatePickers/DatePicker/DatePicker.composables.ts
pioziol123: https://github.com/bethinkpl/design-system/pull/424/commits/34b73202ebfb54c0dd0f90856e89ab6272b0ebb1#diff-e9034fcf709028b88fcba617f264fd901170a4fce563706e72450b3248e43db0R42

lib/js/components/DatePickers/DatePicker/DatePicker.vue
```diff
@@ -134,200 +135,136 @@ import {
...
+	}
+});
+
+const tileIcon = computed(() => {
+	if (!props.icon) {
+		return null;
+	}
+
+	return isOpen.value ? ICONS.FA_CHEVRON_UP : props.icon;
+});
```
tweetgeek, line 269: IMO: we should definitely have some linterns for composition files - it's a terrible mess - I will add this on retro

lib/js/components/DatePickers/DateRangePicker/DateRangePicker.stories.ts
```diff
@@ -0,0 +1,146 @@
...
+	},
+} as ArgTypes;
+
+Interactive.argTypes = argTypes;
+Interactive.args = args;
+
+Interactive.parameters = {
+	design: {
+		type: 'figma',
+		url: 'https://www.figma.com/design/03ABNCSDYWYDmOPJOBGM5l/INI-153-Planowanie-pracy-z-lekcjami?node-id=246-205008&',
```
tweetgeek, line 144:
```suggestion
		url: 'https://www.figma.com/design/izQdYyiBR1GQgFkaOIfIJI/LMS---DS-Components?node-id=8502-1209&t=mHbWSeRfkspainMZ-4',
```

lib/js/components/DatePickers/DatePicker/DatePicker.vue
pioziol123: We should propose something for retro. As far as I know Igor and Damian will ask about what we introduce since we are the first creating component in composition api

pioziol123: We can start with https://eslint.vuejs.org/rules/

lib/js/components/DatePickers/DateRangePicker/DateRangePicker.stories.ts
```diff
@@ -0,0 +1,146 @@
+import DateRangePicker from './DateRangePicker.vue';
```
tweetgeek, line 1: i think that we decide that new stories should be written in new api template like in https://github.com/bethinkpl/design-system/blob/7c73389e632d2788018bb262f2af5e443ffcb7aa/lib/js/components/Well/Well.stories.ts#L2-L1

I can't find any official confirmation, so I'll leave it as **OPTIONAL** and I'll add this on retro

lib/js/components/DatePickers/DatePicker/DatePicker.vue
pioziol123: Sorry, it is for options api

lib/js/components/DatePickers/DatePicker/DatePicker.composables.ts
rogatty: @pioziol123 add delay inside `createDatePicker()` and see if the UX is acceptable or is everything hanging. Maybe we need a loading spinner somewhere?

Then block the flatpickr chunk from being loaded and see if the UX is acceptable. Maybe we need an alert or a dedicated state?

lib/js/components/DatePickers/DateRangePicker/DateRangePicker.stories.ts
```diff
@@ -0,0 +1,152 @@
...
+export default meta;
+
+type Story = StoryObj<DateRangePickerProps>;
+
+const now = Date.now();
+const oneDayMili = 86400000;
+export const Interactive: Story = {
+	args: {
+		isInteractive: true,
+		placeholder: 'Wybierz datę',
```
tweetgeek, line 138:
```suggestion
		placeholder: 'Ustaw',
```

tweetgeek (approved):

tweetgeek:
@pioziol123 we should add this component to export from DS

PR #427, opened 2024-10-17
IT-6019 Fix hover state on IconButton on mobile

fraszczakszymon:


lib/js/components/Buttons/IconButton/IconButton.vue
```diff
@@ -289,6 +292,7 @@ export default defineComponent({
 		WnlIcon,
 		WnlButton,
 	},
+	mixins: [hoverState],
```
rogatty, line 295: Let's not use mixins anymore, we can do better now. 🙏 
It's a good opportunity to talk with @pioziol123 about composability conventions, see https://github.com/bethinkpl/design-system/pull/424#discussion_r1796570861.

lib/js/composables/hover-state.ts
```diff
@@ -0,0 +1,38 @@
+import { Ref, ref } from 'vue';
+import { isTouchDevice } from '../utils/device';
+
+interface HoverState {
+	hovered: Ref<boolean>;
```
rogatty, line 5: `isHovered`
https://bethink.atlassian.net/wiki/spaces/IT/pages/1367539761/Konwencje+techniczne#:~:text=Nazwy%20p%C3%B3l%20typu%20boolean%20powinny%20zaczyna%C4%87%20si%C4%99%20od%20is/are/has/have/can/should.

lib/js/components/Tooltip/Tooltip.vue
```diff
@@ -54,7 +55,7 @@ export default defineComponent({
 		tooltipParams() {
 			return {
 				position: this.placement,
-				event: this.isTouchDevice() ? 'focus' : 'hover',
+				event: isTouchDevice() ? 'focus' : 'hover',
```
rogatty, line 58: We can now remove https://github.com/bethinkpl/design-system/blob/36c10f571453bc4926dc1c2074d384445ca0e811/lib/js/components/Tooltip/Tooltip.vue#L83-L91

lib/js/composables/hover-state.ts
```diff
@@ -0,0 +1,38 @@
...
+	touchEnd: () => void;
+}
+
+export function useHoverState(): HoverState {
+	const hovered = ref(false);
+
+	return {
+		hovered,
+		mouseOver() {
+			if (!isTouchDevice()) {
```
rogatty, line 18: We can cache the result of this function, I don't think it's possible to change in runtime.

rogatty (approved):

PR #431, opened 2024-10-28
Stop propagation on chip remove click

fraszczakszymon:
In case of need to use `@click` on DsChip, `@remove` event should stop propagation


fraszczakszymon:
Ten był dobry, ale chyba niepotrzebny 😄

PR #433, opened 2024-10-30
fix meta width, fix hex border

tweetgeek:


lib/js/components/RichList/RichListItem/RichListItem.vue
```diff
@@ -610,6 +615,7 @@ export default defineComponent({
 				'-ds-verticalWithMedia': this.isVertical && this.hasMedia,
 				'-ds-vertical': this.isVertical,
 				'-ds-loading': this.state === RICH_LIST_ITEM_STATE.LOADING,
+				'-ds-drag': this.state === RICH_LIST_ITEM_STATE.DRAG,
```
pioziol123, line 618: Should we somehow limit this state? e.x. when component is not dragable or is not interactive or maybe other rules?

lib/js/components/RichList/RichListItem/RichListItem.consts.ts
```diff
@@ -7,6 +7,7 @@ export type RichListItemType = typeof RICH_LIST_ITEM_TYPE[keyof typeof RICH_LIST
 export const RICH_LIST_ITEM_STATE = {
 	DEFAULT: 'default',
 	LOADING: 'loading',
+	DRAG: 'drag',
```
pioziol123, line 10: BasicRichListItem changes content colors when is in hover. Look at textGroupState computed in BasicRLI

pioziol123:
Something is no yes in BasicRLI. The bug is already on prod, can you fix it as well? 

https://github.com/user-attachments/assets/1643f87e-1483-46c6-9c05-02a0589b44ac

PR #457, opened 2025-02-04
add max width prop to tile additional text

tweetgeek:


pioziol123 (approved):

lib/js/components/DatePickers/DatePicker/DatePicker.vue
```diff
@@ -240,6 +240,10 @@ export default defineComponent({
 			type: Date,
 			default: null,
 		},
+		updatePositionBasedOnScrollableSelector: {
```
pioziol123, line 243: Should we add it to DateRangePicker as well?

PR #458, opened 2025-02-10
IT-5869 date picker box is public

pioziol123:


lib/js/components/DatePickers/DateRangePicker/DateRangePicker.vue
```diff
@@ -1,17 +1,15 @@
...
-			:end-date="endDateIfDifferentThanStartDate"
+			:end-date="endDate"
 			:start-icon="startIcon"
 			:end-icon="endIcon"
 			:are-icons-hidden-on-mobile="areIconsHiddenOnMobile"
 			:state="state"
 			:color="color"
-			:start-date-eyebrow-text="eyebrowText"
-			:end-date-eyebrow-text="endDateEyebrowText"
 			:is-open="isOpen"
```
tweetgeek, line 13: should we rename it to `is-hoverable` or something like that? when we pass isOpen we only change background color on hover and icon color

pioziol123: We actually block the change of a background color to hover specific.

lib/js/components/DatePickers/DateBox/DateBox.stories.ts
```diff
@@ -0,0 +1,100 @@
...
+				return args;
+			},
+			data() {
+				return {
+					ICONS: Object.freeze(ICONS),
+				};
+			},
+			computed: {
+				formattedStartDate() {
+					console.debug(this.startDate);
```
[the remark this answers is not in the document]
pioziol123: https://github.com/bethinkpl/design-system/pull/458/files#diff-e1ce717e1179a0db14e90ec5374768a206651ca807db58352991eb7895ed7c9cR23

lib/js/components/DatePickers/DateRangePicker/DateRangePicker.vue
tweetgeek: but the name `is open` means that something is open in this component - in this situation we do not open anything, we only turn hovering on or off - so the name is misleading

pioziol123: `is-hoverable` would mean that hover state is completely disabled or enabled, but the pointer is always the same. I can change it on more meaningful name

PR #459, opened 2025-02-19
IT-6172 | Avatar component

latata:
⚠️ Update TS version in DS requires some updates in `wnl-platform`: https://github.com/bethinkpl/wnl-platform/pull/5471

TODO:
* update default size value after UX decision


lib/js/components/Avatar/Avatar.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+				username,
+			},
+		});
+
+		const hexToRgb = (hex: string) => {
+			const result = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex);
+
+			if (!result) {
+				throw new Error('Invalid hex color');
+			}
```
wbrychcy, line 121: I'm not sure if we need in tests error handler if we have under control input, wdyt?

lib/js/components/Avatar/Avatar.vue
```diff
@@ -0,0 +1,178 @@
+<template>
+	<div
+		class="ds-avatar"
+		:class="{
```
wbrychcy, line 4: Maybe can move this logic to computed property?

lib/js/components/Avatar/Avatar.spec.ts
latata: it's only to make TS happy ;)

lib/js/components/Avatar/Avatar.vue
latata: I like it defined here so all the possible class name are visible in the template

wbrychcy (approved):

PR #467, opened 2025-03-11
update picker when date change

tweetgeek:


lib/js/components/DatePickers/DatePicker/DatePicker.composables.ts
```diff
@@ -130,6 +131,21 @@ export function initFlatpickr({
 			flush: 'post', // Ensure updates happen after DOM changes
 		},
 	);
+	watch(
+		[() => props.date],
```
pioziol123, line 135: What about `startDate` and `endDate` in DateRangePicker? Is there a similar scenario? Maybe we should add it to the DatePickerComposablesProps interface as well

lib/js/components/DatePickers/DatePicker/DatePicker.composables.ts
```diff
@@ -130,6 +133,25 @@ export function initFlatpickr({
...
 		},
 	);
+	watch(
+		[() => props.date, () => props.startDate, () => props.endDate],
+		() => {
+			if (props.date) {
+				datePicker?.setDate(props.date, false);
+				datePicker?.updateValue(false);
+				datePicker?.jumpToDate(props.date, false);
+			} else if (props.startDate && props.endDate) {
```
pioziol123, line 143: It seems to work correctly
```suggestion
			} else if (props.startDate || props.endDate) {
```

lib/js/components/DatePickers/DatePicker/DatePicker.composables.ts
```diff
@@ -130,6 +139,23 @@ export function initFlatpickr({
...
 	);
+	watch(
+		[() => props.date, () => props.startDate, () => props.endDate],
+		() => {
+			if (props.date) {
+				updateDatePickerDates(props.date);
+			} else if (props.startDate && props.endDate) {
+				updateDatePickerDates([props.startDate, props.endDate]);
+			} else if (props.startDate && !props.endDate) {
+				updateDatePickerDates(props.startDate);
```
pioziol123, line 150:
```suggestion
				updateDatePickerDates(props.startDate);
			 } else if (!props.startDate && props.endDate) {
				updateDatePickerDates(props.endDate);
```

pioziol123 (approved):

PR #468, opened 2025-03-11
IT-6222 | Badge component

latata:


.eslintrc.cjs
```diff
@@ -2,7 +2,7 @@ module.exports = {
 	plugins: ['import', '@typescript-eslint'],
 	extends: [
 		'eslint:recommended',
-		'plugin:vue/vue3-recommended',
+		'plugin:vue/recommended',
```
latata, line 5: updated `eslint-plugin-vue` so it accepts props destructing

.eslintrc.cjs
```diff
@@ -105,7 +105,6 @@ module.exports = {
 		'vue/no-useless-v-bind': 'error',
 		'vue/padding-line-between-blocks': 'error',
 		'vue/require-direct-export': 'error',
-		'vue/v-on-function-call': 'error',
```
latata, line 108: this rule does not exist anymore in new version of the eslint plugin

PR #478, opened 2025-03-26
IT-6210 fix icons and content spacings in rli

pioziol123:


lib/js/components/RichList/RichListItem/RichListItem.vue
```diff
@@ -74,6 +74,230 @@
 	</div>
 </template>
 
+<script lang="ts">
```
pioziol123, line 77: Don't know why this changed, but it seems lint:fix does that to all components

tweetgeek (approved):

PR #479, opened 2025-04-01
IT-6250 | Add flex-shrink to main avatar element

t-tomalak:
We don't want this element to shrink, therefore, let's explicitly set the flex-shrink to 0 so the element is not affected by other around it.

### DIFF
--- a/lib/js/components/Avatar/Avatar.vue
+++ b/lib/js/components/Avatar/Avatar.vue
@@ -47,6 +47,8 @@
 	$root: &;
 
 	display: flex;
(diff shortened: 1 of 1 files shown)


lib/js/components/Avatar/Avatar.vue
```diff
@@ -47,6 +47,7 @@
 	$root: &;
 
 	display: flex;
+	flex-shrink: 0;
```
latata, line 50: please leave comment in the code 🙏

t-tomalak: added

latata (approved):

PR #483, opened 2025-05-13
create flatpickr instance after toggle

tweetgeek:


lib/js/components/DatePickers/DateRangePicker/DateRangePicker.vue
```diff
@@ -134,6 +135,10 @@ export default defineComponent({
 			type: String,
 			default: '',
 		},
+		createInstanceOnToggle: {
```
tweetgeek, line 138: @pioziol123 do you have an idea for a better name?

lib/js/components/DatePickers/DatePicker/DatePicker.vue
```diff
@@ -272,15 +285,25 @@ export default defineComponent({
...
+					if (!flatpickrInstance.value) {
+						flatpickrInstance.value = (await createDatePicker(
+							flatpickrInputRef.value,
+							datePickerRef.value,
+							props.updatePositionBasedOnScrollableSelector,
+						)) as DatePickerInstance;
+					}
+				}
+			},
+			{ flush: 'post' },
```
pioziol123, line 305: OOC: Why flush post?

lib/js/components/DatePickers/DatePicker/DatePicker.vue
```diff
@@ -272,15 +285,25 @@ export default defineComponent({
...
-		});
+		watch(
+			[() => props.isInteractive, () => props.state, () => props.createInstanceOnToggle],
+			async () => {
+				if (
+					props.isInteractive &&
+					props.state === DATE_PICKER_STATES.DEFAULT &&
+					!props.createInstanceOnToggle
+				) {
+					if (!flatpickrInstance.value) {
```
pioziol123, line 296: ifs can be merged

lib/js/components/DatePickers/DatePicker/DatePicker.vue
```diff
@@ -140,10 +140,23 @@
...
 
-import DsTile, { TILE_ADDITIONAL_TEXT_MAX_WIDTHS, TILE_BORDER_COLORS } from '../../Tile';
+import { Instance as DatePickerInstance } from 'flatpickr/dist/types/instance';
+
 import { IconItem, ICONS } from '../../Icons/Icon';
+import DsTile, {
+	TILE_ADDITIONAL_TEXT_MAX_WIDTHS,
+	TILE_BORDER_COLORS,
+	TileColors,
+	TileStates,
```
pioziol123, line 150:
```suggestion
```
Seems to be unused

lib/js/components/DatePickers/DateRangePicker/DateRangePicker.vue
pioziol123: What you think about making create on toggle the only behavior? It would simplify code a lot. Maybe we should even destroy picker when it's not used?

lib/js/components/TextGroup/TextGroup.vue
```diff
@@ -266,6 +266,7 @@ $text-group-colors: (
 <script lang="ts">
 import { defineComponent, PropType } from 'vue';
 import DsSkeleton from '../Skeleton/Skeleton.vue';
+import DsTooltip from '../Tooltip';
```
pioziol123, line 269: Are those changes in imports something dictated by our linters or your configuration? They add some noise to PRs and I don't really want to check them if it is not necessary, and I have to verify them not to miss actual change in imports.

lib/js/components/DatePickers/DatePicker/DatePicker.vue
tweetgeek: there are used, in template. https://github.com/bethinkpl/design-system/pull/483/files#diff-a929dcc8ab9e69d0279737dcd6b2516b873ea1e3aa78723dce333a4cdfdd6cecR27

lib/js/components/TextGroup/TextGroup.vue
tweetgeek: cursor does some shit to organize imports and sorts them alphabetically by default, I previously had a change to this file but it was reverted and this import remained. fixed

pioziol123: I speak more generally because import reorganization is something that happens a lot in your PRs

lib/js/components/DatePickers/DatePicker/DatePicker.vue
```diff
@@ -273,12 +281,16 @@ export default defineComponent({
...
 		});
 		watch([() => props.isInteractive, () => props.state], async () => {
-			if (props.isInteractive && props.state === DATE_PICKER_STATES.DEFAULT) {
-				await createDatePicker(
+			if (
+				props.isInteractive &&
+				props.state === DATE_PICKER_STATES.DEFAULT &&
+				!flatpickrInstance.value
+			) {
+				flatpickrInstance.value = (await createDatePicker(
```
pioziol123, line 289: Is this watch necessary? We don't want to create watch on state change, rather on toggle. Also, there is a difference in assignment. Here you assign an instance to the value property, below you assign to the variable itself

lib/js/components/TextGroup/TextGroup.vue
tweetgeek: i know
> cursor does some shit to organize imports and sorts them alphabetically by default

fixed - i mean that i disabled this rule, but I don't know if it's good, Damian always pushed for alphabetical order - it's easier to read.

lib/js/components/DatePickers/DatePicker/DatePicker.vue
pioziol123: We still leave flatpickr when we close it. Maybe we should create/destroy it on toggle and then, probably, we don't need this watcher.

lib/js/components/DatePickers/DateRangePicker/DateRangePicker.vue
```diff
@@ -158,6 +158,11 @@ export default defineComponent({
 			emit('update:date', { startDate: event[0], endDate: event[1] });
 		};
 
+		const onClose = () => {
+			destroyDatePicker();
```
pioziol123, line 162: You don't have to import this method. You can call `destroyDatePicker` in composable. The only thing would be this ref cleanup which is good enough for me

lib/js/components/DatePickers/DatePicker/DatePicker.vue
```diff
@@ -279,15 +284,10 @@ export default defineComponent({
 		} = initFlatpickr({
 			props,
 			onChange,
+			onClose,
 			defaultDates: props.date ?? new Date(),
 			mode: 'single',
 		});
-		watch([() => props.isInteractive, () => props.state], async () => {
```
pioziol123, line 275: Removing this watcher created a weird scenario. 
1. Open a date picker
2. Set `isInteractive` prop to false
3. You can still select a date. I don't know if this is a problem, tho. Let it be for now.

pioziol123 (approved):

PR #484, opened 2025-05-15
IT-6386 | add teleport to modal

latata:


lib/js/components/Modals/Modal/Modal.vue
```diff
@@ -1,105 +1,107 @@
...
-							</div>
-							<wnl-button
-								v-if="footerTertiaryButtonText"
-								:type="BUTTON_TYPES.TEXT"
-								:color="BUTTON_COLORS.NEUTRAL"
-								:icon-left="footerTertiaryButtonIcon"
-								:state="footerTertiaryButtonState"
-								class="ds-modal__tertiaryButton"
-								@click="$emit('tertiary-button-click')"
+	<teleport to="body">
```
latata, line 2: Just wrapped with teleport

kvas-damian: > Just wrapped with teleport - no more changes in this PR

+ `v-bind="$attrs"` and `inheritAttrs: false,`

lib/js/components/Modals/Modal/Modal.vue
kvas-damian, line 1: Don't we need the same change to https://github.com/bethinkpl/design-system/blob/7678764349e603e5f8266b7f4d83c9dd5bd5dd10/lib/js/components/Modal/Modal.vue to avoid similar problems (not verified yet if it's present in our code now). Modal has to be rendered inside another stacking context that has lower z-index than nav to observe this bug.

latata: I couldn't find a problematic use case for the old modal, but I added teleport to it and tested it with two existing use cases on the platform (IsUncertainFeatureOnboardingModal, MultistepProblemHelpModal - in overlay) - they work fine.

PR #485, opened 2025-05-19
IT-6387 - add Chip inside Well component

tomaszperkowski:


lib/js/components/Well/Well.stories.ts
```diff
@@ -2,20 +2,46 @@
 
 import { Args, ArgTypes, Meta, StoryObj } from '@storybook/vue3';
 import Well from './Well.vue';
-import { WELL_PADDINGS } from './Well.consts';
+import Banner from './Banner.vue';
```
t-tomalak, line 5: This seems to be unused

lib/js/components/Well/Well.vue
```diff
@@ -26,29 +40,41 @@
...
+		right: 16px;
+		top: -10px;
+	}
 }
 </style>
 
-<script>
+<script lang="ts" setup>
 import { WELL_PADDINGS } from './Well.consts';
+import Chip, { CHIP_COLORS, CHIP_DEFAULT_COLOR, CHIP_RADIUSES } from '../Chip/index.ts';
```
t-tomalak, line 54: I don't think that `/index.ts` is needed

lib/js/components/Well/Well.vue
```diff
@@ -26,29 +40,41 @@
...
+import { ICONS } from '../Icons/Icon';
+import type { Value } from '../../utils/type.utils';
+
+type Padding = keyof typeof WELL_PADDINGS;
```
t-tomalak, line 58: We probably should move this to the const file

lib/js/components/Well/Well.vue
```diff
@@ -26,29 +40,41 @@
 	&.-ds-small {
 		padding: $space-xs;
 	}
+
+	&__chipContainer {
+		position: absolute;
+		right: 16px;
```
t-tomalak, line 46: `rights` value should be equal to the padding of the content

lib/js/components/Well/Well.vue
```diff
@@ -26,29 +40,41 @@
...
+	hasChip = false,
+	chipLabel = null,
+	chipLabelUppercase = false,
+	chipLeftIcon = null,
+	chipRadius = CHIP_RADIUSES.ROUNDED,
+	chipColor = CHIP_DEFAULT_COLOR,
+	chipColorHex = null,
+} = defineProps<{
+	padding?: Padding;
+	hasChip?: boolean;
```
t-tomalak, line 72: We probably should have validation for this value together with padding, or maybe at least have the chip not rendered so have some sort of compute that will check if padding is set and return this value accordingly.

lib/js/components/Well/Well.vue
```diff
@@ -6,6 +6,19 @@
...
+		<div v-if="hasChip" class="ds-well__chipContainer">
+			<chip
+				:label="chipLabel"
+				:is-label-uppercase="chipLabelUppercase"
+				:left-icon="chipLeftIcon"
+				:radius="chipRadius"
+				:color="chipColor"
+				:color-hex="chipColorHex"
+			>
+				<slot name="chipAccessory" />
```
t-tomalak, line 18: I don't quite understand the slot in here? I don't think it can be used in any way? If we want to expose the accessory we would need to wrap slot in template with the correct name and then expose slot in it

lib/js/components/Well/Well.vue
```diff
@@ -26,29 +40,41 @@
...
+	chipRadius = CHIP_RADIUSES.ROUNDED,
+	chipColor = CHIP_DEFAULT_COLOR,
+	chipColorHex = null,
+} = defineProps<{
+	padding?: Padding;
+	hasChip?: boolean;
+	chipLabel?: string;
+	chipLabelUppercase?: boolean;
+	chipLeftIcon?: IconItem;
+	chipRadius?: Value<typeof CHIP_RADIUSES>;
```
t-tomalak, line 76: Those types could be defined in the Chip consts file

t-tomalak (dismissed):
When checking the Storybook chip without ICON seems to be positioned too low, and overall it seems to be moving up and down when rendering with ICON or without it

lib/js/components/Well/Well.vue
tomaszperkowski: we discussed it on daily, we dont make such exclusions in components

tomaszperkowski:
@t-tomalak I checked how the ‘Chip’ component behaves, and the issue lies within it. I couldn’t quickly see what exactly is wrong with it.

lib/js/components/Well/Well.vue
```diff
@@ -18,37 +33,54 @@
...
+	chipLeftIcon = null,
+	chipRadius = CHIP_RADIUSES.ROUNDED,
+	chipColor = CHIP_DEFAULT_COLOR,
+	chipColorHex = null,
+} = defineProps<{
+	padding?: WellPadding;
+	hasChip?: boolean;
+	chipLabel?: string;
+	chipLabelUppercase?: boolean;
+	chipLeftIcon?: IconKey;
```
t-tomalak, line 81: I think this should be `IconItem` type

lib/js/components/Chip/Chip.consts.ts
```diff
@@ -24,3 +28,5 @@ export const CHIP_RADIUSES = {
 	CAPSULE: 'capsule',
 	ROUNDED: 'rounded',
 } as const;
+
+export type ChipRadius = keyof typeof CHIP_RADIUSES;
```
t-tomalak, line 32: Those seems to be incorect types as those are returning keys `CAPSULE` | `ROUNDED` and instead we should use `Value` helper `Value<typeof CHIP_RADIUSES>` to return values of const

lib/js/components/Well/Well.consts.ts
```diff
@@ -2,3 +2,10 @@ export const WELL_PADDINGS = {
 	SMALL: 'small',
 	MEDIUM: 'medium',
 };
+
+export type WellPadding = keyof typeof WELL_PADDINGS;
```
t-tomalak, line 6: Same here it should be `Value`

t-tomalak (commented):
@tomaszperkowski so what we are doing with the "jumping" chip? We cannot proceed with it like that as is not matching the designs as chip without icon is definitely too low

lib/js/components/Well/Well.consts.ts
```diff
@@ -1,4 +1,13 @@
...
+
 export const WELL_PADDINGS = {
 	SMALL: 'small',
 	MEDIUM: 'medium',
 };
+
+export type WellPadding = Value<typeof WELL_PADDINGS>;
+
+export const CHIP_AVAILABLE_PADDINGS = {
+	SMALL: 'small',
```
bognix, line 11: I wonder if this should be coupled with `WELL_PADDINGS`, now it's not, but at the end of the day values in this const are the same values as in `WELL_PADDINGS`.

t-tomalak: This is used in the "stories" only so yeah, I think we can drop it and just "combine" the array in there from the values of WellPaddings .

lib/js/components/Well/Well.stories.ts
```diff
@@ -2,20 +2,45 @@
...
 		setup() {
 			return { args };
 		},
-		template: '<well v-bind="args"><div v-html="args.content" /></well>',
+		template: wrapWithContainer(
+			'<well v-bind="args" :chip-left-icon="ICONS[args.chipLeftIcon]"><div v-html="args.content" /></well>',
+		),
+		data() {
+			return {
+				ICONS: Object.freeze(ICONS),
```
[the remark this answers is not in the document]
t-tomalak: I probably does make sense to not use it here either, so will move it to setup

bognix (approved):

PR #495, opened 2025-06-04
add solid reference icons

preziu:


lib/js/icons/fontawesome.ts
```diff
@@ -233,7 +235,7 @@ export const FONTAWESOME_ICONS = {
 	FA_ARROWS_LEFT_RIGHT_SOLID: fasArrowsLeftRight,
 	FA_ARROWS_MAXIMIZE: faArrowsMaximize,
 	FA_ARROWS_ROTATE: faArrowsRotate,
-	FA_ATOM: faAtom,
```
rogatty, line 236: This is a backward incompatible change, there will be errors on production until we stop using the old icon. I suggest the following:
1. Add `FA_ATOM_SOLID` and others, merge to platform
2. Update icons used on prod, make sure `FA_ATOM` is no longer used
3. Update https://github.com/bethinkpl/wnl-platform/blob/c16cd3a2709deb809d6324b689fcb2d16f908fb1/resources/assets/js/modules/multistepProblems/components/MultistepProblemPartReferencesEntryPoint.spec.ts#L25
4. Remove `FA_ATOM` and release as major version (see https://semver.org/lang/pl/), because it's a breaking change

PR #497, opened 2025-06-05
IT-6455 | Card: new props: footerHasPadding, paddingSize

latata:


lib/js/components/Buttons/Button/Button.vue
```diff
@@ -37,129 +38,77 @@
...
+	color = BUTTON_COLORS.PRIMARY,
+	radius = BUTTON_RADIUSES.CAPSULE,
+	state = BUTTON_STATES.DEFAULT,
+	iconLeft = null,
+	iconRight = null,
+	elevation = BUTTON_ELEVATIONS.NONE,
+	as = 'span',
+} = defineProps<{
+	// Fixme - remove `string` from types - it is used only to avoid breaking changes for now
+	size?: ButtonSize | string;
```
kvas-damian, line 81: why do we need to union these types with `string`?

lib/js/components/Buttons/Button/Button.vue
```diff
@@ -37,129 +38,77 @@
...
+	type = BUTTON_TYPES.FILLED,
+	color = BUTTON_COLORS.PRIMARY,
+	radius = BUTTON_RADIUSES.CAPSULE,
+	state = BUTTON_STATES.DEFAULT,
+	iconLeft = null,
+	iconRight = null,
+	elevation = BUTTON_ELEVATIONS.NONE,
+	as = 'span',
+} = defineProps<{
+	// Fixme - remove `string` from types - it is used only to avoid breaking changes for now
```
kvas-damian, line 80: did you identify such cases, where we used wrong prop value?

latata: not wrong value, but a value typed as string which caused TS errors

lib/js/components/Cards/Card/Card.vue
```diff
@@ -59,32 +82,50 @@
 	}
 
 	&.-ds-border-size-small {
-		border-radius: $radius-xs;
 		border-top-width: $border-s;
 
+		#{$root}:not(.-ds-flat) & {
```
kvas-damian, line 87: there are leftovers of solution based on prop `isFlat`

lib/js/components/Banner/Banner.vue
```diff
@@ -169,20 +221,25 @@
...
-		padding: 0 $space-4xs;
+		padding: 0 $space-2xs;
 		width: 100%;
 	}
 
 	&__title {
 		@include heading-s-default-bold;
 
-		color: $color-neutral-text-strong;
+		align-items: center;
```
kvas-damian, line 231: does it change anything while `__iconWrapperSmall` has `align-self: stretch;`?

lib/js/components/Banner/Banner.vue
```diff
@@ -86,64 +110,73 @@
...
-				}
+			#{$self}__textWrapper {
+				flex-direction: row;
+			}
 
-				#{$self}__textWrapper {
-					flex-direction: row;
-				}
+			#{$self}__rightWrapper {
+				padding: 0 0 0 $space-s;
```
kvas-damian, line 133: this spacing is visible even when right slot is empty. Did you consider migrating it to `gap`?

lib/js/components/Banner/Banner.vue
```diff
@@ -86,64 +110,73 @@
...
+				padding: $space-xs 0;
 
-				#{$self}__buttonWrapper {
-					padding: $space-xs 0;
-				}
-
-				#{$self}__rightSlot {
-					flex-grow: initial;
-					padding: $space-xs 0 $space-xs 0;
+				.-ds-small & {
```
kvas-damian, line 140: this selector is invalid. Is this block required?

lib/js/components/Banner/Banner.vue
```diff
@@ -1,20 +1,44 @@
 <template>
 	<div class="ds-banner">
-		<div class="ds-banner__inner" :class="[colorClass, layoutClass]">
+		<div
+			class="ds-banner__inner"
+			:class="[
```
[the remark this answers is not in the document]
latata: It was like that before. And IMO it makes sense as we treat `.ds-banner` only as container to handle `@container` query.

lib/js/components/Banner/Banner.vue
```diff
@@ -86,64 +110,73 @@
...
+			}
 		}
 
 		&.-ds-fail {
 			background-color: $color-fail-background;
 			border-color: $color-fail-border-weak;
+
+			&.-ds-titleInColor {
+				--ds-banner-title-color: #{$color-fail-text};
+			}
```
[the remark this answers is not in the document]
latata: The benefit is that it's selector-agnostic ;) If we change the title element class name it would still work ;)

lib/js/components/Banner/Banner.vue
latata: Yes, the position of the title text is different when I remove it

lib/js/components/Banner/Banner.spec.ts
```diff
@@ -0,0 +1,148 @@
+import { describe, it, expect, vi } from 'vitest';
+import { mount } from '@vue/test-utils';
+import Banner from './Banner.vue';
+import { BANNER_COLORS, BANNER_SIZES } from './Banner.consts';
+import { ICONS } from '../Icons/Icon';
+import { h } from 'vue';
+
+describe('Banner', () => {
```
tweetgeek, line 8: optional: I asked copilot about possible additional tests

Here are some additional test scenarios you can consider adding to further strengthen coverage for the `Banner` component:

### 1. **Button Click Event**
- Test that clicking the CTA button emits the `button-clicked` event.
- If the button is not rendered (no `buttonText` prop), verify the event is not emitted.

### 2. **Close Button Visibility**
- Test that the close (`closable`) button is not rendered when `closable` is false or missing.

### 3. **Right Slot Priority**
- If both `buttonText` and `rightSlot` are provided, ensure both are rendered and their order is correct.

### 4. **Icon Hidden on Mobile**
- When `isIconHiddenOnMobile` is true, check the appropriate class is applied to hide the icon on mobile.

### 5. **Slot Fallbacks**
- If neither `defaultText` nor `expandedText` slots are provided, ensure the component renders gracefully without errors.

### 6. **Title in Color**
- When `titleInColor` is true, check that the `-ds-titleInColor` class is applied to the inner element and the style is updated.

### 7. **Reactivity of `isExpanded` Prop**
- If the `isExpanded` prop is changed externally, ensure the component reacts correctly and updates its expanded state.

### 8. **Prop Validation**
- Try passing invalid values for `color` or `size` and check that the component either renders the default or fails gracefully.

### 9. **Accessibility**
- Ensure that the banner, buttons, and interactive elements have appropriate ARIA roles and attributes.

### 10. **Multiple Banners**
- Render multiple `Banner` components simultaneously to ensure no state leakage or cross-interaction.

tweetgeek (approved):

lib/js/components/Banner/Banner.spec.ts
latata: Some of them would be valuable for sure and I would be happy to add them. But how about in a separate PR so I can merge this one sooner? ;)

PR #500, opened 2025-06-10
IT-6466 | Medium modal width increased to 800px, added rightActions slot to modal

latata:


lib/js/components/Modals/Modal/Modal.vue
```diff
@@ -179,10 +181,10 @@ $image-height-small: 140px;
 	}
 
 	&__content {
-		padding: $space-l $space-s;
+		padding: $space-xl $space-s;
```
kvas-damian, line 184: According to Figma, bottom padding shouldn't be changed

lib/js/components/Modals/Modal/Modal.stories.ts
```diff
@@ -24,7 +24,7 @@ const StoryTemplate: StoryFn<typeof Modal> = (args) => ({
 		':footerTertiaryButtonText="footerTertiaryButtonText" :footerTertiaryButtonIcon="ICONS[footerTertiaryButtonIcon]" :footerTertiaryButtonState="footerTertiaryButtonState" ' +
 		':footerCheckboxText="footerCheckboxText" :headerImage="headerImage">' +
 		'<div v-html="defaultSlot" />' +
-		'<template #rightActions><div style="display: inline-flex;" v-html="rightActionsSlot" /></template>' +
+		'<template #rightActions><div style="display: inline-flex;" v-if="rightActionsSlot" v-html="rightActionsSlot" /></template>' +
```
kvas-damian, line 27: if we want to disable this slot when `rightActionsSlot` is empty, then it's better to move condition to the `<template>`.
It doesn't matter here becasue there is no additional HTML markup for the slot, but in Vue empty slot is different than no slot

kvas-damian (approved):

PR #503, opened 2025-06-17
IT-6419 | FormField updates after DR

latata:


lib/js/components/Form/FormField/FormField.vue
```diff
@@ -83,13 +82,11 @@
...
 		flex: 1 0 0;
 		gap: $space-5xs;
-		min-height: 28px;
 	}
 
 	&__label {
-		align-items: baseline;
-		display: flex;
-		gap: $space-5xs;
+		line-height: $typography-line-height-2xs;
```
kvas-damian, line 88: Instead of using a random variable that matches the value, I'd suggest using `@include formLabel-m-default-bold;`. Then you could remove this mixin from `__labelText`

kvas-damian (approved):

PR #514, opened 2025-07-29
IT-6541 | Add small variants to Tile component

t-tomalak:
In Tile, we want to have option to render it slightly different depending on the size of tile:
* Show only left icon
* Show only right icon
* Hide right icon - this was available through separate property "isRightIconHiddenOnMobile"

As container-type is applying some "different" styles to the element and changing it behaviour we need to only apply that when querying container size is really needed.
Also, we have been using "margins" for left and right icon which is unnecessary and instead we should use gap, which will make it easier to center elements for small variants.
DatePicker seems to be the only user of previous isRightIconHiddenOnMobile, so we should change the functionality to the `smallVariant` and use the `RIGHT_ICON_HIDDEN`

### DIFF
--- a/lib/js/components/DatePickers/DatePicker/DatePicker.spec.ts
+++ b/lib/js/components/DatePickers/DatePicker/DatePicker.spec.ts
@@ -9,7 +9,7 @@ const createComponent = (props = {}) => {
 	});
 };
 
-describe('DrawerHeader', () => {
+describe('DatePicker', () => {
 	describe('Hidden Icon', () => {
 		it('should pass DEFAULT compact layout to Tile when isIconHiddenOnMobile is not set', () => {
 			const component = createComponent();


lib/js/components/Tile/Tile.consts.ts
```diff
@@ -46,3 +46,12 @@ export const TILE_ADDITIONAL_TEXT_MAX_WIDTHS = {
 } as const;
 
 export type TileAdditionalTextMaxWidth = Value<typeof TILE_ADDITIONAL_TEXT_MAX_WIDTHS>;
+
+export const TILE_COMPACT_LAYOUTS = {
```
bognix, line 50: why the name `TILE_COMPACT_LAYOUTS` - if I understand correctly, this is the list of all layouts (because `default` is present on this list), not only the compact ones.

lib/js/components/Tile/Tile.vue
```diff
@@ -370,68 +406,58 @@ $tile-colors: (
...
+const tileCompactLayoutClass = computed(() => {
+	switch (props.compactLayout) {
+		case TILE_COMPACT_LAYOUTS.ICON_LEFT_ONLY:
+			return '-ds-iconLeftOnly';
+		case TILE_COMPACT_LAYOUTS.ICON_RIGHT_ONLY:
+			return '-ds-iconRightOnly';
+		case TILE_COMPACT_LAYOUTS.ICON_RIGHT_HIDDEN:
+			return '-ds-iconRightHidden';
+		default:
+			return false;
```
bognix, line 437: I wonder if that would work, but perhaps if you return `''` instead of `false` you could remove the `!!` casting in template and use `tileCompactLayoutClass` directly in `:class=[...]` binding.

lib/js/components/Tile/Tile.vue
```diff
@@ -318,6 +320,40 @@ $tile-colors: (
...
+			display: none;
+		}
+	}
+
+	&.-ds-onlyLeftIcon,
+	&.-ds-onlyRightIcon,
+	&.-ds-rightIconHidden {
+		container-type: inline-size;
+
+		@container (width > 165px) {
```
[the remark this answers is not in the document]
bognix: @pioziol123 are you satisfied with the answer? Can we resolve this conversation?

bognix (approved):
minor comments, looks good to me.

lib/js/components/Tile/Tile.consts.ts
bognix: discussed offline - we are leaving the names as they are.

lib/js/components/DatePickers/DatePicker/DatePicker.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+import DatePicker from './DatePicker.vue';
+import { Tile, TILE_COMPACT_LAYOUTS } from '../../../';
+
+const createComponent = (props = {}) => {
+	return mount(DatePicker, {
+		props,
+	});
+};
+
+describe('DrawerHeader', () => {
```
bognix, line 12:
```suggestion
describe('DatePicker', () => {
```

PR #583, opened 2026-04-01
IT-7404 | SelectListItem accessory slot, replace SelectListItemTile with SelectListItemContainer, TextGroup XS size

latata:
## Summary

- **SelectListItem**: migrated from Options API to `<script setup>`, added named `accessory` slot
- **SelectListItemContainer**: new generic slot-based container replacing `SelectListItemTile`
- **TextGroup**: added `x-small` size variant
- **Divider**: added `flex-shrink: 0` to prevent shrinking in flex containers
- **SelectListItemToggle**: added `PropType` to `size` and `state` props

## Breaking Changes

`SelectListItemTile` has been removed. Migrate by replacing prop forwarding with a slot child:

```vue
<!-- Before -->
<ds-select-list-item-tile
  :text="text"
  :eyebrow-text="eyebrowText"
  :state="state"
/>

<!-- After -->
<ds-select-list-item-container>
  <ds-tile
    :text="text"
    :eyebrow-text="eyebrowText"
    :state="state"
  />
</ds-select-list-item-container>
```


lib/js/components/SelectList/SelectListItemContainer/SelectListItemContainer.vue
```diff
@@ -0,0 +1,19 @@
+<template>
```
tweetgeek, line 1: should we add `barrel export` as index.ts to this component?

lib/js/components/SelectList/SelectListItemContainer/SelectListItemContainer.stories.ts
```diff
@@ -0,0 +1,72 @@
+import SelectListItemContainer from './SelectListItemContainer.vue';
+import Tile from '../../Tile';
+import { TILE_STATES } from '../../Tile';
```
tweetgeek, line 3:
```suggestion
import Tile, { TILE_STATES } from '../../Tile';
```

lib/js/components/SelectList/SelectListItem/SelectListItem.vue
```diff
@@ -169,75 +172,28 @@ import {
...
+
+const {
+	iconLeft = null,
+	isSelected = false,
+	label,
+	eyebrowText = '',
+	isEyebrowTextUppercase = false,
+	selectionMode = SELECT_LIST_ITEM_SELECTION_MODE.SELECT_ONLY,
+	size = SELECT_LIST_ITEM_SIZES.SMALL,
+	state = SELECT_LIST_ITEM_STATES.DEFAULT,
```
tweetgeek, line 185: Is it okay that we're removing validators?

latata: TS is now the validator ;)

lib/js/components/SelectList/SelectListItemContainer/SelectListItemContainer.vue
latata: Other SelectList* components don't have - so let's be consistent ;)

lib/js/components/SelectList/SelectListItem/SelectListItem.vue
tweetgeek: We validate this during TS compilation, but runtime validation isn't working right now

latata: I know, this is what's current approach is. Every time we migrate from Options API to Composition API we get rid of validators. I believe runtime validators are unnecessary complication when we have TS

tweetgeek (approved):

kvas-damian (approved):

PR #586, opened 2026-04-16
BT-184 | translate survey placeholders

kvas-damian:
https://bethink.atlassian.net/browse/BT-184

### DIFF
--- a/lib/js/i18n/en/survey.json
+++ b/lib/js/i18n/en/survey.json
@@ -1,0 +1,3 @@
+{
+	"ds.survey.openEnded.placeholder": "Enter your answer"
+}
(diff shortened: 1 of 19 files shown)


lib/js/i18n/en/survey.json
```diff
@@ -0,0 +1,3 @@
+{
+	"ds.survey.openEnded.placeholder": "Enter your answer"
```
latata: Consider renaming this key to `ds.survey.placeholder` — the `openEnded` segment is too specific since the same key is reused in `SurveyQuestionScale` and `SurveyQuestionTextarea`, neither of which are "open-ended" components. A generic name would better reflect the shared usage.

> 🤖 Generated with [Claude Code](https://claude.ai/claude-code)

latata (approved):
Remember to update i18n files in the platform code: https://github.com/bethinkpl/wnl-platform/blob/master/resources/assets/js/i18n/pl/index.ts#L70 and `en/index.ts`

PR #590, opened 2026-04-22
fix(IconButton): increase specificity to prevent ds-button style override

latata:


lib/js/components/Buttons/IconButton/IconButton.vue
```diff
@@ -148,7 +148,8 @@
 		color: map-get($icon-button-colors, 'theme', 'hovered');
 	}
 
-	&__button {
+	// .ds-button increases specificity to ensure base button styles never override icon button styles
+	&__button.ds-button {
```
kvas-damian: we use internal `Button` class name. I expect we won't remember to update this place when we change button class name. Is it possible to make this selector more specific without using Button internals?

kvas-damian (approved):

PR #601, opened 2026-05-28
feat(dark-theme): add dark theme support

latata:
## Summary

- Adds a `.-ds-dark` CSS class-based dark theme that can be applied to `<html>` to switch all design system components to dark mode
- Introduces per-brand dark raw color scales (`_raw-wnl-dark`, `_raw-mc-dark`, `_raw-bodywork-dark`) that invert the `--theme-*` palette for dark contexts
- Ships a full 513-token dark semantic color override file (`_tokens-variables-dark.scss`) covering all `--color-*` tokens under `.-ds-dark`
- Adds `--raw-gray-950` and `--raw-pure-black` raw color tokens used as dark-mode base surfaces
- Switches scrollbar mixin defaults from hardcoded `white` to `var(--color-default-background, white)` so scrollbar tracks adapt automatically to the active color scheme
- Adds dark-mode flatpickr theme scoped under `.-ds-dark` in both `DatePicker` and `DateRangePicker`
- Adds a Storybook toolbar for toggling color scheme (light/dark) and brand (`wnl`, `mc`, `bodywork`) via a `decorators` export that applies `.-ds-dark` to `<html>` and brand classes to `<body>`

## Test plan

- [ ] Toggle dark mode in Storybook and verify components render correctly in all three brands
- [ ] Verify flatpickr calendar renders in dark mode when `.-ds-dark` is on `<html>`
- [ ] Verify page-level scrollbar adapts in dark mode on Chrome/Edge (WebKit)
- [ ] Check Firefox — scrollbar color is WebKit-only; standard `scrollbar-color` property is not yet in the mixin
- [ ] Confirm `.contrastBackground` wrapper in Button/IconButton/CounterToggle stories shows correct contrast in both modes


lib/styles/mixins/_scrollbars.scss
```diff
@@ -1,4 +1,8 @@
-@mixin scrollbar($size, $foreground-color, $background-color: white) {
+@mixin scrollbar(
+	$size,
+	$foreground-color,
+	$background-color: var(--color-default-background, white)
```
kvas-damian: why do we keep `white` as a fallback value? Should we have a fallback everywhere?

latata: Not necessary, let me remove it

lib/styles/design-system.scss
kvas-damian: did you consider using color tokens? `color-neutral-background-hovered` should work here. Then we could remove separate styles for dark theme

lib/styles/settings/colors/_tokens-variables-dark.scss
```diff
@@ -0,0 +1,513 @@
+.-ds-dark {
```
kvas-damian, line 1: are these colors imported from Figma by script? If not, maybe it's the right moment to apply them in Figma, export them and modify our script to handle that?

kvas-damian (approved):

PR #603, opened 2026-06-10
new token: color-default-background-rgb

latata:


tools/importers/ImportColorsFromFigmaTokens.ts
```diff
@@ -166,6 +181,20 @@ export const ImportTokens = (
 				label: tokenName,
 				value: resolvedValue,
 			});
+
+			if (TOKENS_WITH_RGB_VARIANT.has(tokenName)) {
+				const rgbTokenName = `${tokenName}-rgb`;
+				const rgbValue = toRgbReference(resolvedValue);
```
kvas-damian: can't we use `parseTokenReference(tokenEntry.$value + '-rgb')`?

kvas-damian (approved):

tools/importers/ImportColorsFromFigmaTokens.ts
latata: Good call — done in 35c83d8a. `parseTokenReference(tokenEntry.$value + '-rgb')` produces the same `var(--raw-…-rgb)` output, so I dropped the bespoke `toRgbReference` helper entirely.

PR #605, opened 2026-06-11
IT-7544 | Add magic color variant to Button and IconButton

latata:
## What

Adds a new `magic` color variant to `Button` and `IconButton`, rendering a multi-stop gradient (fill, ghost background, and angular/conic border treatments).

## Changes

**Button**
- New `MAGIC` button color variant.
- `useMagicGradient` composable that injects the SVG gradient `<def>` into the DOM so icon/spinner fills can reference it via `url(#...)`, with a dedicated disabled-state gradient.
- Magic fill driven through the `--ds-magic-fill` CSS custom property.

**IconButton**
- New `magic` color variant, including ghost hovered/focused backgrounds for the icon-only outlined style (no glow shadow in icon-only mode).
- Refactored state handling to forward the full `state` to the underlying `Button` via `ICON_BUTTON_STATE_MAP` instead of deriving hovered-only state.

**Styles**
- New `lib/styles/settings/_gradients.scss` with magic gradient tokens (linear ghost backgrounds, conic borders, disabled variants).
- New magic color variables/definitions under `lib/styles/settings/colors/`.

## Tests

- Added specs for `useMagicGradient`, `Button`, and `IconButton` covering the magic variant.


lib/styles/components/_buttons.scss
```diff
@@ -201,6 +202,89 @@ $button-disabled-alpha: 0.6;
...
+		&:hover,
+		&.-ds-hovered,
+		&:active,
+		&.-ds-active {
+			// Re-declare the full stack so the gradient ring survives, with the ghost tint
+			// layered on top.
+			background:
+				$gradient-magic-background-ghost-hovered padding-box,
+				linear-gradient($color-default-background, $color-default-background) padding-box,
+				$gradient-magic-border-angular border-box;
```
tweetgeek: Should we add a mixin for this? It's repeated several times with different $gradient-magic-background.

tweetgeek (approved):

lib/styles/components/_buttons.scss
latata: Good call — extracted a `setMagicOutlinedBackground($tint, $border)` mixin and routed all four states (base, hover, focus, disabled) through it. Done in bc3abeaa.

tweetgeek (approved):

PR #608, opened 2026-06-12
feat(IconButton,Chip): add layout-neutral touch hit area via ::after on pointer:coarse

latata:
## Problem

The old approach to touch targets used `min-height`/`min-width: 40px` (the `-ds-touchable` class), which inflated the layout footprint of every IconButton by default — pushing surrounding elements apart even on desktop.

## Solution

Add a `touchableHitArea` SCSS mixin (`lib/styles/mixins/_touchable.scss`) that extends the tap zone to a minimum of 32px on touch devices (`@media (pointer: coarse)`) using a transparent `::after` pseudo-element. The visual element keeps its real size; only the invisible hit area grows.

## Changes

- **New mixin** `touchableHitArea` — absolutely-positioned `::after` centered on the host, sized `max(100%, 32px)`. Requires the host to have `position: relative`.
- **IconButton** — `@include touchableHitArea` applied unconditionally; `position: relative` added to root. Old `touchable` prop renamed to `touchableDeprecated` (default changed `true` → `false`) to flag it for removal. The `-ds-touchable` layout-inflating class path is kept but deprecated.
- **Chip** — `@include touchableHitArea` + `position: relative` added to `.-ds-interactive` modifier.
- **Call sites** — removed explicit `:touchable="false"` overrides that are now redundant (new default is `false`). Preserved `touchable-deprecated` on sites that still need the 40px layout inflation (OverlayHeader, Pagination, Modals close buttons).
- **`$min-touchable-size`** renamed to `$DEPRECATED-min-touchable-size` in `_icons.scss`.
- **HelpButton story** — removed stale localization warning banner.


lib/js/components/Icons/Icon/Icon.vue
```diff
@@ -55,7 +55,12 @@ export default defineComponent({
 			type: Boolean,
 			default: false,
 		},
-		touchable: {
+		/**
+		 * @deprecated Adds the `-ds-touchable` class, which enlarges the hit area to the
+		 * deprecated `$DEPRECATED-min-touchable-size`. Still has effect; renamed to flag it
+		 * for removal so call sites can migrate away from this layout-inflating behavior.
+		 */
+		touchableDeprecated: {
```
kvas-damian, line 63: Do I understand correctly that we don't add new touchable here because `Icon` isn't meant to be a clickable element?

kvas-damian (approved):

lib/styles/mixins/_touchable.scss
```diff
@@ -0,0 +1,19 @@
+$touchable-size: 32px;
+
+// Enlarge the touch target on touch devices without affecting the layout footprint.
+// The visual element keeps its real size; a transparent ::after extends the hit area.
+// NOTE: the host element must establish a positioning context (e.g. `position: relative`)
```
kvas-damian, line 5: did you consider adding `position: relative` inside a mixin?

lib/js/components/Icons/Icon/Icon.vue
latata: yup

lib/styles/mixins/_touchable.scss
latata: I considered that, but Claude convinced me to keep it outside to avoid potential duplication. Also, it doesn't always need to be `position: relative`.

PR #609, opened 2026-06-12
Merge build.yml and storybook.yml into publish-npm.yml

latata:
## Context

Until now `build.yml` and `storybook.yml` both ran on **every push to `master`**:

- **`build.yml`** installed deps, ran `yarn build` (vite → `dist/`), and committed the built `dist/` back to `master` via `JamesIves/github-pages-deploy-action` (the recurring `Deploying to master … 🚀` commits). `dist/` is a **git-tracked** folder, and `publish-npm.yml` (manual `workflow_dispatch`) relied on `dist/` already being present in `master` — it bumped the version, pushed, created a release, and ran `npm publish` **without building anything**.
- **`storybook.yml`** ran `yarn storybook:build` and deployed the static Storybook to the git-tracked `docs/` folder on `master` (served by GitHub Pages), again via `JamesIves/github-pages-deploy-action`.

This means the package **and** the Storybook site were rebuilt and redeployed on every merge, even when no release was intended.

## Change

Fold both the dist build and the Storybook build into `publish-npm.yml` and delete `build.yml` and `storybook.yml`, so build/deploy happens **only when we explicitly publish a package**, not on every merge to `master`.

### `publish-npm.yml`
- Added `FONT_AWESOME_AUTH_TOKEN` job env (needed by the build).
- Added the `node_modules` cache step.
- Added an **Install and Build** step before the version bump that now runs `yarn install` + `yarn build` + `yarn storybook:build:docs` + `touch ./docs/.nojekyll`.
- Version bump now uses `npm version --no-git-tag-version`, then `git add -A` → **single commit** containing the freshly built `dist/`, the regenerated `docs/` (Storybook), and the bumped `package.json`, plus an annotated `vX.Y.Z` tag. One commit, one push via the existing `--follow-tags`.
- GitHub release and `npm publish` steps unchanged.

**Final step order:** checkout (`fetch-depth: 0`) → setup-node (registry-url) → cache → install+build (dist + storybook) → git identity → bump version + commit dist/docs + tag → push → GH release → `npm publish`.

### `package.json`
- Added a `storybook:build:docs` script (`storybook build -c .storybook -o docs`) that builds Storybook straight into the tracked `docs/` folder. The existing `storybook:build` (→ `public/storybook`) is left untouched — it's still used by `deploy_storybook_to_s3.sh`, `ci.yml`, and `build-to-review-on-trigger.yml`.

### `build.yml` / `storybook.yml`
- Both deleted. They were the only things building/deploying `dist/` and `docs/` on push to `master`; nothing else references them.

## Why the `github-pages-deploy-action` steps were dropped

The old deploys used `JamesIves/github-pages-deploy-action@v4`, which creates its **own separate commit** (`Deploying to master … 🚀`) and pushes it independently. Keeping them was incompatible with the goal here:

- They would produce **multiple commits** (dist, docs, version bump), not the single commit requested.
- They would mean **multiple separate pushes to `master`** in one run — a non-fast-forward collision risk.

Replacing them with `git add -A && git commit` stages the built `dist/` and regenerated `docs/` alongside the bumped `package.json` into one commit, pushed once.

### Behavioral notes
- **dist:** the old action used `clean: false`, so it **never deleted** stale files in `master`'s `dist/`. Since vite emits hash-named files and wipes `dist/` on each build (`emptyOutDir` defaults to `true`), `master` accumulated orphaned artifacts over time. With `git add -A`, deletions are staged too, so each publish commit yields a **clean `dist/`**. An improvement, but a change from the prior additive behavior.
- **docs:** `storybook build` empties its output dir before writing, so stale/removed stories are dropped (matching the old `clean: true`). `.nojekyll` is re-created (storybook wipes it) so GitHub Pages keeps serving the site without Jekyll processing.
- `docs/` is **not** published to npm (the `files` field ships only `dist/`, `lib/`, `package.json`, `README.md`), so committing it adds no package bloat.

## Verification
- [ ] Trigger **Publish to NPM** via `workflow_dispatch` (e.g. `patch`) and confirm: a single new `master` commit with the version bump + updated `dist/` + regenerated `docs/`, a matching `vX.Y.Z` tag and GitHub release, GitHub Pages serving the updated Storybook, and an npm tarball containing the freshly built `dist/`.


.github/workflows/build.yml
```diff
@@ -1,33 +0,0 @@
...
-      - uses: actions/cache@v4
-        name: Cache node_modules
-        id: cache-node-modules
-        with:
-          path: node_modules
-          key: design-system-node-modules-${{ hashFiles('yarn.lock') }}
-      - name: Install and Build 🔧
-        run: | # Install npm packages and build the Storybook files
-          yarn install --immutable --immutable-cache
-          yarn build
```
radekbenkel, line 26: question(non-blocking): Shouldn't we build then as part of `ci.yaml`? If I'm not mistaken, if we somehow introduce some change that will break the build we will know about it only when someone will try to release it, as there no `yarn build` during PR checks.

.github/workflows/publish-npm.yml
```diff
@@ -32,15 +34,30 @@ jobs:
...
           node-version-file: 'package.json'
           registry-url: 'https://registry.npmjs.org'
+      - uses: actions/cache@v5
+        name: Cache node_modules
+        id: cache-node-modules
+        with:
+          path: node_modules
+          key: design-system-node-modules-${{ hashFiles('yarn.lock') }}
+      - name: Install and Build 🔧
+        run: | # Install npm packages, build dist and Storybook
```
radekbenkel, line 44: question: Shouldn't this run AFTER version bumping? Assuming that we're releasing next patch version from `v1.0.0`, this will be built with `v1.0.0` while `bump` step will update `package.json` to `v1.0.1`.

.github/workflows/build.yml
latata: Good point, though I'd prefer not to tackle it as part of this PR

.github/workflows/publish-npm.yml
latata: not sure what's the difference? 🤔

latata: I think version in package.json  does not impact the build at all

radekbenkel: If something would rely on this value it would not be correct (one release off). But I don't see were have such cases right now (no Sentry or similar tracking `package.json` version), so I think we can close this as is.

radekbenkel (approved):

PR #616, opened 2026-06-24
Switch npm publish from NPM_TOKEN to OIDC Trusted Publishing

radekbenkel:
## Summary

- Replaces the long-lived `NPM_TOKEN` secret with npm OIDC Trusted Publishing — GitHub Actions mints a short-lived identity token on demand, no stored credential needed
- Adds `id-token: write` permission required for OIDC
- Upgrades npm to latest before publish (OIDC requires npm ≥ 11.5.1, which may not be bundled with the Node versions we support)
- Drops `NODE_AUTH_TOKEN` / `NPM_TOKEN` from the publish step
- Adds `--provenance` to attach a public build attestation linking the published package to this repo and CI run

The Trusted Publisher is already configured on npmjs.com for `bethinkpl/design-system → publish-npm.yml` so this is ready to merge and test.

### DIFF
--- a/.github/workflows/publish-npm.yml
+++ b/.github/workflows/publish-npm.yml
@@ -24,6 +24,7 @@ jobs:
     if: github.ref == 'refs/heads/master'
     permissions:
       contents: write
+      id-token: write
     env:
       FONT_AWESOME_AUTH_TOKEN: ${{ secrets.FONT_AWESOME_AUTH_TOKEN }}
     steps:
@@ -34,6 +35,8 @@ jobs:
         with:
           node-version-file: 'package.json'
           registry-url: 'https://registry.npmjs.org'
+      - name: Upgrade npm for Trusted Publishing
+        run: npm install -g npm@latest
       - uses: actions/cache@v5
         name: Cache node_modules
         id: cache-node-modules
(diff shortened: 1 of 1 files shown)


latata (approved):

.github/workflows/publish-npm.yml
```diff
@@ -34,6 +35,8 @@ jobs:
         with:
           node-version-file: 'package.json'
           registry-url: 'https://registry.npmjs.org'
+      - name: Upgrade npm for Trusted Publishing
+        run: npm install -g npm@latest
```
rogatty, line 39: That's pretty brave. How confident are we in npm never shipping malware in the latest version or making breaking changes? Can we use a specific version that supports OIDC instead? I'm not pushing, just always looking for problems. 🙂

radekbenkel: `npm` supply chain attack in `npm` itself would be next level of irony 😅 

Would hardcoding it to `v11.17.0` for now work for you https://socket.dev/npm/package/npm/overview/11.17.0? 

In the future after switching to Node LTS we will use one bundled with Node.

latata: 🙄 We're feeling brave! It's only temporary, and we'll get rid of it once we upgrade to Node 24.

rogatty: I'd hardcode it but can live with it being dynamic temporarily if you have plans to get rid of it soon.

PR #621, opened 2026-06-26
Update Node/Yarn to LTS and harden supply-chain config

radekbenkel:
## Summary

Upgrades the toolchain to the current Node/Yarn LTS line and hardens the install against supply-chain risks.

### Toolchain
- **Node** pinned to `^24.5.0` via `engines.node` as the single source of truth — read by `actions/setup-node` in CI (and by version managers that resolve `engines`), resolving the range to the latest matching 24.x.
- **Yarn** bumped `4.11.0` → `4.17.0`; lockfile refreshed (format `v8` → `v10`).
- **npm** floor raised to `>=11.5.1`. The CI `npm install -g npm@…` step is removed — Node 24.5+ already bundles npm ≥ 11.5.1, the minimum for OIDC Trusted Publishing (`npm publish --provenance`).
- CI workflows updated to Node 24 and bumped to current GitHub Actions (`checkout@v6`, `setup-node@v6`, `cache@v5`, `github-script@v7`), all reading `node-version-file: 'package.json'`.

### Supply-chain hardening (`.yarnrc.yml`)
- `enableScripts: false` — lifecycle scripts disabled globally, re-enabled only for `esbuild` and `vue-demi` via `dependenciesMeta.built`.
- `npmMinimalAgeGate: "1w"` — guards against fast-flux / freshly-published malicious versions.
- `approvedGitRepositories` restricted to the single git dependency (`vue-popper`).

## Notes
- `engines.node: ^24.5.0` is advisory for consumers — older-Node apps get an `EBADENGINE` warning, not a hard block (unless they opt into `engine-strict`). The shipped dist runs in the browser via the consumer's bundler, so consumer Node version is an install-time signal only.
- No separate `.nvmrc`/`.node-version` added: `engines.node` is the single pin for both CI and local development, avoiding a second source of truth to keep in sync.


latata:
forgot to give ✅  🙈

PR #637, opened 2026-08-03
IT-7662: Ecommerce changes

latata:
Ongoing branch for the IT-7662 ecommerce component changes. More component work will be added here, so this stays a draft until the batch is complete.

## SelectField

New `DsSelectField` (`lib/js/components/Form/SelectField/`) — a single-select field built on reka-ui's `Select` primitives, wrapped in `FormField` so it shares the label / sub-label / help / message / field-status anatomy with the other form fields.

- `options` accepts a flat list, option groups (`{ label?, options }`) or a mix of both. `normalizeOptions` collapses consecutive ungrouped options into one unlabelled group and keeps the original order; a labelled group renders as a real `role="group"`, an unlabelled one renders its options as direct children of the listbox (an unlabelled reka `SelectGroup` would emit `role="group"` with a dangling `aria-labelledby`). A divider is inserted between groups.
- Options are rendered through the existing `SelectList` / `SelectListItem` / `SelectListSectionTitle` / `SelectListItemDivider` components, so the dropdown matches the rest of the design system. Each option supports `iconLeft`, `eyebrowText` and `isDisabled`.
- Props: `placeholder`, `leftIcon`, `ariaLabel` (used when there is no visible label), `maxHeight` (number → px, string → any CSS length; defaults to 320px and is always clamped to the space available in the viewport), plus `v-model` and an `update:open` event.
- vee-validate: passing `name` binds the field to the surrounding `useForm()` context via the new `useSelectFieldWithinForm`. A select has no `input` event, so closing the listbox stands in for blur and marks the field touched; validation errors surface as the field message and the error state.
- Trigger styling mirrors `InputField` (inset shadow, border colors for hover / focus / open / error / disabled) with the chevron rotating while open. The listbox is portalled to `<body>`, so `.ds-selectField__content` is styled from an unscoped block (same approach as `DatePicker.vue`), widens to the longest option but never below the trigger width, and scrolls past `maxHeight`.
- Stories cover the interactive playground, groups, icons + eyebrow text, a long list, long labels in a constrained width, disabled and error. `SelectField` was also added to the `Form/SimpleForm` story so the vee-validate wiring is visible next to the other fields.

## Card: elevation / radius / background / padding rework

Aligns Card with the redesigned Figma component, where the three visual axes are independent variants instead of being bundled into `isFlat`.

- New `elevation` prop (`default` | `top` | `none`) replacing the shadow part of `isFlat`. `top` casts the shadow upwards, for cards docked to the bottom of the viewport.
- `hasRadius` becomes `radius` (`all` | `top` | `none`), with `top` rounding only the upper corners.
- `backgroundColor` gains a `none` value, so the old `isFlat` look is composed from `elevation="none"` + `radius="none"` + `backgroundColor="none"`.
- New `medium` padding size (24px) between `small` (16px) and `large` (32px), applied across header, divider, content and footer.

Ribbon corner clipping now follows `radius` alone. A left ribbon spans the full card height, so its bottom corner is only rounded when the card is rounded at the bottom, and the ribbon's own radius applies only at `radius="none"`.

## Foundations: Shadows/Top tokens

Adds `$shadow-top-s/m/l/xl` from [Figma foundations](https://www.figma.com/design/CG4lkRM9SnX7Qjz6rZxLYZ/LMS---DS-Foundations?node-id=209-2868) — each mirrors its `$shadow-*` counterpart with negated Y offsets, reusing the theme-aware shadow colour tokens — plus the matching `foundations/Shadows` story entries. `elevation="top"` consumes `$shadow-top-s`.

## Supporting changes

- **`lib/styles/mixins/_dropdown-surface.scss`** — the floating-panel surface (background, `$shadow-m`, `overflow: hidden`) and its per-edge radius mixins, extracted so `DsDropdown` and SelectField's listbox stay in sync. `Dropdown.vue` now consumes the mixins instead of repeating the declarations; no visual change there.
- **`SelectListItem`** gains a `text` slot (falling back to `label`) so SelectField can wrap the visible label in reka's `SelectItemText`, which is what `aria-labelledby` and the trigger's value display point at. `label` is still used for the `title` attribute. Also adds the component's first spec file.
- **`vitest.setup.ts`** stubs `ResizeObserver`, which jsdom lacks and floating-ui's `autoUpdate` (used by every reka popper-based component) requires.
- **`lib/js/index.ts`** exports `DsSelectField` and the SelectField types.

## Breaking changes

- `isFlat` and `hasRadius` are removed. Both are kept as `RemovedProp` markers, so leftover usages fail `ts:check` with a migration hint and stay hidden from Storybook controls.
- The internal modifier classes changed: `-ds-flat` and `-ds-noRadius` are gone, replaced by `-ds-elevationTop` / `-ds-elevationNone`, `-ds-radiusTop` / `-ds-radiusNone` and `-ds-backgroundNone`. Relevant to any app targeting those classes in its own CSS.

No in-repo consumer passed `isFlat`, `hasRadius` or `paddingSize`, so no internal migration was needed. `CardExpandable` does not forward the new props (out of scope).

## Verification

`yarn format:check`, `yarn test` (560/560), `yarn ts:check` and `yarn lint` all clean. Visual pass in Storybook still pending review: the four new shadow tiles in light and dark, the `none`/`none`/`none` combo against the old `isFlat` look, `radius="top"` with both ribbon positions, and the SelectField stories (trigger states, grouped list, long list scrolling, long-label ellipsis).

### DIFF
--- a/lib/js/components/Form/SelectField/SelectField.vue
+++ b/lib/js/components/Form/SelectField/SelectField.vue
@@ -1,0 +1,328 @@
+<template>
+	<form-field v-bind="formFieldProps">
+		<template #field="{ fieldId, messageId }">
+			<select-root
+				v-model="value"
+				:disabled="isDisabled"
+				:required="formFieldProps.hasRequiredIndicator"
+				@update:open="onOpenChange"
+			>
+				<select-trigger
+					:id="fieldId"
+					:aria-describedby="hasMessage ? messageId : undefined"
+					:aria-label="formFieldProps.label ? undefined : ariaLabel"
+					:class="[
+						'ds-selectField__trigger',
+						{
+							'-ds-error': formFieldProps.state === FORM_FIELD_STATES.ERROR,
+							'-ds-disabled': isDisabled,
+						},
+					]"
+				>
+					<ds-icon
+						v-if="leftIcon"
+						class="ds-selectField__leftIcon"
+						:icon="leftIcon"
+						:size="ICON_SIZES.X_SMALL"
+					/>
+					<select-value class="ds-selectField__value" :placeholder="placeholder" />
+					<select-icon as-child>
+						<ds-icon
+							class="ds-selectField__icon"
+							:icon="ICONS.FA_CHEVRON_DOWN"
+							:size="ICON_SIZES.X_SMALL"
+						/>
+					</select-icon>
+				</select-trigger>
+
+				<select-portal>
+					<select-content
+						class="ds-selectField__content"
+						position="popper"
+						align="start"
+						:side-offset="0"
+						:style="contentStyle"
+					>
+						<select-viewport as-child>
+							<ds-select-list>
+								<template
+									v-for="(group, groupIndex) in normalizedOptions"
+									:key="groupIndex"
+								>
+									<select-separator v-if="groupIndex > 0" as-child>
+										<ds-select-list-item-divider />
+									</select-separator>
+
+									<!-- A labelled group gets real `role="group"` semantics. -->
+									<select-group v-if="group.label">
+										<select-label as-child>
+											<ds-select-list-section-title :label="group.label" />
+										</select-label>
+										<select-field-option
+											v-for="option in group.options"
+											:key="option.value"
+											:option="option"
+											:is-selected="option.value === value"
+										/>
+									</select-group>
+
+									<!--
+										Ungrouped options render as direct children of the listbox:
+										an unlabelled SelectGroup would emit `role="group"` with a
+										dangling `aria-labelledby`.
+									-->
+									<template v-else>
+										<select-field-option
+											v-for="option in group.options"
+											:key="option.value"
+											:option="option"
+											:is-selected="option.value === value"
+										/>
+									</template>
+								</template>
+							</ds-select-list>
+						</select-viewport>
+					</select-content>
+				</select-portal>
+			</select-root>
+		</template>
+		<!-- begin: FormField slots -->
+		<template v-if="$slots.help" #help>
+			<slot name="help" />
+		</template>
+		<template v-if="$slots.labelAside" #labelAside>
+			<slot name="labelAside" />
+		</template>
+		<template v-if="$slots.message" #message="{ messageId }">
+			<slot name="message" :message-id="messageId" />
+		</template>
+		<template v-if="$slots.fieldStatus" #fieldStatus>
+			<slot name="fieldStatus" />
+		</template>
+		<!-- end: FormField slots -->
+	</form-field>
+</template>
+
+<style lang="scss" scoped>
+@import '../../../../styles/settings/spacings';
+@import '../../../../styles/settings/radiuses';
+@import '../../../../styles/settings/borders';
+@import '../../../../styles/settings/colors/tokens';
+@import '../../../../styles/settings/typography/tokens';
+@import '../../../../styles/settings/shadows';
+@import '../../../../styles/settings/animations';
+
+.ds-selectField {
+	$root: &;
+
+	// The element rules come before `&__trigger` so that the trigger's state overrides, which
+	// target the same elements at a higher specificity, stay in ascending order.
+	&__leftIcon {
+		color: $color-neutral-icon-weak;
+	}
+
+	&__value {
+		@include formText-s-default-regular;
+
+		color: $color-neutral-text-heavy;
+		flex: 1;
+		// override the default min-width so long values can ellipsize
+		min-width: 0;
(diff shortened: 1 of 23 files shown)


lib/js/components/Form/SelectField/SelectField.vue
```diff
@@ -0,0 +1,328 @@
...
+import { assertOptionValues, normalizeOptions, toCssLength } from './SelectField.utils';
+import { useSelectFieldWithinForm } from './useSelectFieldWithinForm';
+
+const { options, placeholder, leftIcon, ariaLabel, maxHeight, name, ...rest } =
+	defineProps<SelectFieldProps>();
+
+const slots = defineSlots<SelectFieldSlots>();
+
+const emit = defineEmits<{
+	'update:open': [isOpen: boolean];
```
tweetgeek: `'update:open'` declares a `v-model:open` update event, but there is no corresponding `open` prop and nothing is forwarded to `SelectRoot`'s `open`/`defaultOpen`, so the modifier is write-only.

lib/js/components/Form/SelectField/SelectField.vue
```diff
@@ -0,0 +1,328 @@
...
+	max-height: min(
+		var(--select-field-max-height, 320px),
+		var(--reka-select-content-available-height)
+	);
+	// Widens to fit the longest option but never narrower than the trigger, and never wider
+	// than the space available in the viewport.
+	max-width: var(--reka-select-content-available-width);
+	min-width: var(--reka-select-trigger-width);
+	// Portalled to <body>, so it has to clear modals
+	z-index: $z-index-modal + 1;
```
tweetgeek: minor: `z-index: $z-index-modal + 1` evaluates to 16777271, which is both the Safari maximum documented in `lib/styles/settings/_z-indexes.scss:4` and the exact value already claimed by `tooltip: 16777271` in `lib/js/primevue.ts:14`, so the dropdown ties with tooltips and nothing can ever be layered above it

lib/js/components/Form/SelectField/SelectField.vue
```diff
@@ -0,0 +1,328 @@
...
+			color: $color-neutral-text-weak;
+		}
+	}
+
+	&__icon {
+		color: $color-neutral-icon;
+		transition: $default-cubic-bezier-transition;
+	}
+
+	&__trigger {
```
tweetgeek, line 150: this re-implements InputField's control-surface styling (border, radius, inset shadow, hover/focus/error/disabled color rules, left-icon and text colors). Let's move it to a mixin.

lib/js/components/Form/SelectField/useSelectFieldWithinForm.ts
```diff
@@ -0,0 +1,27 @@
...
+ * the listbox stands in for blur — that is when the field is marked touched.
+ *
+ * Revalidation on selection needs no handler here: vee-validate's `useField` defaults to
+ * `validateOnValueUpdate: true`, so changing the value already re-runs validation.
+ */
+export function useSelectFieldWithinForm(
+	name: MaybeRefOrGetter<string | undefined>,
+	modelValue: Ref<SelectFieldValue | undefined>,
+) {
+	const { value, errors, field } = useFormFieldWithinForm<SelectFieldValue>(name, modelValue);
```
tweetgeek: The `useFormFieldWithinForm<SelectFieldValue>` type argument is inert - the shared composable hardcodes `useField<Array<string>>` - so the new select gets no real value typing instead of the shared composable being generalized

lib/js/components/Form/SelectField/SelectField.vue
latata: I'd rather leave this one as is. There are only two surfaces today — InputField's wrapper and SelectField's trigger (PasswordField just composes InputField) — and what they share is a flat list of token declarations. The state rules are exactly where they differ: InputField proxies focus with `:focus-within` because focus lives on the inner `<input>`, while the trigger is focusable itself and also has to look focused while the listbox is open, so it needs `[data-state='open']` + `:focus-visible` — and every `:not()` exclusion in the error/hover rules has to mirror that list. Sharing them would mean passing a selector list into the mixin and building `:not(a):not(b)` chains by interpolation, which reads worse than the two literal blocks. Not enough benefit for the added complication, in my opinion.

lib/js/components/Form/SelectField/SelectField.vue
latata: I set it to `$z-index-modal`
In the future, we should probably revisit our z-index strategy, as using the maximum z-index value isn't a good long-term solution

PR #638, opened 2026-08-05
IT-7536 | Move the DatePickers to a calendar-day API

rogatty:
## What

Every date-shaped prop and emit on `DsDatePicker`, `DsDateRangePicker` and `DsDateBox` now speaks `CalendarDate` — a `'YYYY-MM-DD'` string — instead of `Date`. Adds a `today` prop to both pickers.

**This is a breaking change; the release will be major (42.0.0).** Do not publish before the platform PR is ready — the two migrate in lockstep.

| Member | Was | Becomes |
|---|---|---|
| `DatePicker.date` | `Date`, default `null` | `CalendarDate \| null`, default `null` |
| `DatePicker.minDate` / `.maxDate` | `Date` | `CalendarDate \| null` |
| `DatePicker.disableDates` | `Array<Date>` | `Array<CalendarDate>` |
| `DatePicker.today` | — | **new** `CalendarDate \| null`, default `null` |
| `DatePicker` `update:date` | `(date: Date)` | `(day: CalendarDate)` |
| `DateRangePicker.startDate` / `.endDate` / `.minDate` / `.maxDate` | `Date` | `CalendarDate \| null` |
| `DateRangePicker.disableDates` | `Array<Date>` | `Array<CalendarDate>` |
| `DateRangePicker.today` | — | **new** `CalendarDate \| null`, default `null` |
| `DateRangePicker` `update:date` | `{ startDate: Date; endDate: Date }` | `{ startDate: CalendarDate; endDate: CalendarDate }` |
| `DateBox.startDate` / `.endDate` | `Date`, default `null` | `CalendarDate \| null`, default `null` |

`CalendarDate` is exported from the library root.

## Why

wnl-platform (the only consumer) moved every calendar day to `Y-m-d` strings in bethinkpl/wnl-platform#7907, so it carries a pair of adapter functions at every picker boundary. Two `new Date()` calls also stayed unreachable from the app:

1. flatpickr's `self.now` — it draws the today ring and picks the month an empty calendar opens on. The DS never set `config.now`, so the ring marked *browser* today even when the app's today (the user's profile timezone) was a different day.
2. `DatePicker`'s `defaultDates` fallback — an empty picker opened with browser-now visibly selected.

`today` fixes both: when set, it becomes flatpickr's `now` and the `defaultDates` fallback. Left unset, the pickers behave exactly as before.

A browser-local `Date` now exists only inside the DS, next to the two things that need one: flatpickr and `Intl`. No moment, no timezone library — the DS being zone-free is the design.

## How

`lib/js/components/DatePickers/calendarDate.ts` holds the type, the conversion pair and the prop validators. `parseCalendarDate` uses the numeric constructor deliberately: `new Date('2026-08-05')` parses as UTC midnight, which is the previous day in every browser west of UTC. Write local, read local, and the day survives in every host timezone.

Conversion happens only at the flatpickr boundary — the initial config *and* both watchers — and in the display computeds, which hand the existing `localFullDateWithShortMonthName` / `localWeekdayName` / `localMonthDayWithShortMonthDay` helpers a `Date` in the same zone that wrote it.

Two calls worth a reviewer's eye:

- The range picker keeps `[startDate, endDate]` as its default dates with **no** `today` fallback. In range mode a fallback would invent a selection, and `now` already opens an empty calendar on the right month.
- The `today` watcher assigns `instance.now` as well as calling `set()`, because `set()` only updates the config while the today ring reads the instance's own `now`.

Every date-shaped prop validates `null` or `/^\d{4}-\d{2}-\d{2}$/`, so an ISO instant leaking in fails loudly in dev instead of silently shifting a day.

## If a time-of-day picker is ever needed

Nothing here blocks one — flatpickr supports time natively (`enableTime`, `time_24hr`, `noCalendar` for time-only), and these components still build a browser-local `Date` internally for flatpickr and `Intl`; this PR only changed what crosses the component boundary. A time-capable picker would be a new component (`DsDateTimePicker`) or a new mode with its own emit, not a rework of this API. `DsDatePicker` stays day-only regardless: "a day" and "a moment" are different domain objects, and making that distinction type-visible is the point of this change.

What a time picker forces — and what a `Date` used to hide — is deciding **which zone the picked wall-clock means**. Two coherent contracts exist:

1. **Emit an instant** (a `Date`). Right when the user means their physical wall clock — "notify me at 9:00 where I'm sitting." The picker renders browser-local, the user picks browser-local, the instant is exact.
2. **Emit a zone-free wall-clock string** — `'2026-08-06T09:00'`, a `LocalDateTime` mirroring the `CalendarDate` pattern, anchored by the consumer in whatever zone its domain calls for (wnl-platform: the profile zone).

The trap to document when that day comes: for days, the browser≠profile gap is invisible (any zone's "August 6" is the same day); for times it is the whole offset — the user picks "9:00" on a browser-local widget, and a consumer anchoring it elsewhere stores a different moment than the user's wall clock showed. Option 2 therefore needs the UI to say whose 9:00 it is, or the component needs an explicit zone-offset input so it can render the consumer's zone (still no timezone library here — the consumer passes the offset).

The one thing not to do is widen `CalendarDate` or the existing props to smuggle a time through — that would undo the day/instant distinction this PR establishes.

## Tests

54 new tests across four files: round trip (month boundaries, year boundaries, a leap day), `pl`/`en` display for the picker and the box, emit shape for single and range, `today` reaching flatpickr's `config.now` / `instance.now` / `defaultDate` and browser-now when unset, the prop-watcher path handing flatpickr parsed `Date`s, DateBox single/differing/null days, and the validators.

`yarn test` (561 passed), `yarn ts:check`, `yarn lint` and `yarn build` are green. `dist` is left to the release workflow.

Verified live in Storybook: with `today` a week ahead, the ring sits on 12 August while the browser is on 5 August; picking 18 August round-trips to "Wtorek / 18 sie 2026"; the range picker emits a string pair that renders "10 sie – 20 sie". A new `CustomToday` story demonstrates `today` ≠ browser today.

### DIFF
--- a/lib/js/components/DatePickers/DateBox/DateBox.spec.ts
+++ b/lib/js/components/DatePickers/DateBox/DateBox.spec.ts
@@ -1,0 +1,56 @@
+import { afterEach, describe, expect, it } from 'vitest';
+import { mount } from '@vue/test-utils';
+import { nextTick } from 'vue';
+import DateBox from './DateBox.vue';
+import { SUPPORTED_LOCALE } from '../../../i18n';
+import { setTestLocale } from '../../../tests/helpers';
+
+const createComponent = (props = {}) => mount(DateBox, { props });
+
+const dateTexts = (component: ReturnType<typeof createComponent>) =>
+	component.findAll('.ds-dateBox__dateText').map((element) => element.text());
+
+const eyebrowTexts = (component: ReturnType<typeof createComponent>) =>
+	component.findAll('.ds-dateBox__eyebrow').map((element) => element.text());
+
+describe('DateBox', () => {
+	afterEach(() => {
+		setTestLocale(SUPPORTED_LOCALE.pl);
+	});
+
+	it('should render a single date when the start and the end name the same day', () => {
+		const component = createComponent({ startDate: '2026-08-05', endDate: '2026-08-05' });
+
+		expect(dateTexts(component)).toEqual(['05 sie']);
+		expect(eyebrowTexts(component)).toEqual(['Środa']);
+		expect(component.find('.ds-dateBox__separator').exists()).toBe(false);
+	});
+
+	it('should render both dates when the end names a different day', () => {
+		const component = createComponent({ startDate: '2026-08-05', endDate: '2026-08-17' });
+
+		expect(dateTexts(component)).toEqual(['05 sie', '17 sie']);
+		expect(eyebrowTexts(component)).toEqual(['Środa', 'Poniedziałek']);
+		expect(component.find('.ds-dateBox__separator').exists()).toBe(true);
+	});
+
+	it('should render the placeholder and no eyebrow when both dates are null', () => {
+		const component = createComponent({
+			startDate: null,
+			endDate: null,
+			placeholder: 'Ustaw',
+		});
+
+		expect(dateTexts(component)).toEqual(['Ustaw']);
+		expect(eyebrowTexts(component)).toEqual([]);
+	});
+
+	it('should render the day the string names in the en locale', async () => {
+		const component = createComponent({ startDate: '2026-08-05', endDate: '2026-08-17' });
+		setTestLocale(SUPPORTED_LOCALE.en);
+		await nextTick();
+
+		expect(dateTexts(component)).toEqual(['Aug 05', 'Aug 17']);
+		expect(eyebrowTexts(component)).toEqual(['Wednesday', 'Monday']);
+	});
+});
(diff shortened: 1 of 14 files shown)


rogatty:
This needs to be thought trough again after I've learned about Temporal: https://blog.gaborkoos.com/posts/2026-07-21-Your-JS-Date-Is-Lying-to-You/ and https://www.npmjs.com/package/temporal-polyfill. It might be a better, safer, future-proof alternative approach.

